PR #8, opened 2018-05-23
Implement BareKit

izqui:
`BareKit` will be the kit that the CLI uses in `aragon run` or `dao new` if no template is provided. It creates a DAO and installs the provided app from APM, and assigns all roles.

`KitBase` can also be used by other kits so we can stop duplicating that logic all over.


izqui:
It is now working and there is a branch in the CLI ready for it (https://github.com/aragon/aragon-cli/pull/105)

After this is merged, we can make github.com/aragon/aragen publish kits (atm you need to manually deploy it before you can do 'aragon run')

izqui:
@sohkai I want to merge this soon so it stops being a blocker for aragen kit support :)

PR #16, opened 2018-06-15
[WIP] Implement Payroll Kit

izqui:
Builds on #8

Implements a PayrollKit with permissions ready for a production environment. It grants all admin permissions to a `root` account to be able to upgrade or rescue the funds in case of an incident.  

I think we should build a PayrollDemoKit inside this package too so it can be easy to deploy locally for testing purposes.

Here is a diagram of the current permissions:
2018-06-15 at 7 41 49 pm


facuspagnuolo:
@sohkai @izqui now that we're going to keep only production templates here, should we move these changes to the `PayrollKit` in aragon-apps or just close it?

PR #18, opened 2018-07-25
TCR: Staking deferred lock. Adapt to last app changes

bingen:
- Registry: must be initialized
- Staking: can overlock and defer locks
- Curation: changed redistribution method (see PR #13)


facuspagnuolo:
@bingen @sohkai are we going to maintain this template? If it was an experimental thing we may want to move it into concierge

bingen:
Yes, TCR itself is not finished nor in our immediate roadmap, so I would close all related PRs

PR #25, opened 2018-09-19
Upgrade to aragonOS 4

### LINKED ISSUE #24: Update to aragonOS 4 and the upgraded apps

bingen:
Closes #24 

Needs:
- https://github.com/aragon/aragonOS/pull/428
- https://github.com/aragon/aragon-id/pull/22
- https://github.com/aragon/aragon-id/pull/23

TODO:

- [x] Fix tests
- [x] Reuse KitBare
- [x] Split Democracy and Multisig
- [x] Analyze gas costs
- [x] Update Survey Kit


izqui:
Another TODO for this PR would be to analyze the gas costs

kits/multisig/contracts/MultisigTemplate.sol
```diff
@@ -43,12 +46,18 @@ contract MultisigTemplate is BetaTemplateBase {
...
         );
 
-        uint256 multisigSupport = neededSignatures * 10 ** 18 / signers.length;
+        uint256 multisigSupport;
+        uint256 neededSignaturesE18 = neededSignatures * 10 ** 18;
+        if (neededSignaturesE18 < signers.length) {
+            multisigSupport = 0;
+        } else {
+            multisigSupport = neededSignaturesE18 / signers.length - 1;
+        }
```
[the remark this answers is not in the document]
bingen: Yes, sure, but I was planning to do this in another PR for this [issue](https://github.com/aragon/dao-kits/issues/2) once this is finished.

kits/democracy/arapp.json
```diff
@@ -0,0 +1,6 @@
+{
+  "appName": "democracykit.aragonpm.eth",
```
izqui: `democracy-kit` for consistency

kits/democracy/contracts/DemocracyKit.sol
```diff
@@ -51,5 +53,10 @@ contract DemocracyTemplate is BetaTemplateBase {
             minAcceptanceQuorum,
             voteDuration
         );
+
+        // burn support modification permission
+        ACL acl = ACL(Kernel(voting.kernel()).acl());
+        acl.createPermission(NO_ENTITY, voting, voting.MODIFY_SUPPORT_ROLE(), NO_ENTITY);
```
izqui: Setting the manager to `NO_ENTITY` doesn't burn the permission as they can be re-created when the manager is `0x0`

kits/multisig/arapp.json
```diff
@@ -0,0 +1,6 @@
+{
+  "appName": "multisigkit.aragonpm.eth",
```
izqui: `multisig-kit`

kits/multisig/test/multisig.js
```diff
@@ -74,6 +74,36 @@ contract('Multisig Template', accounts => {
...
+                await voting.changeSupportRequiredPct(multisigSupport.add(1), { from: owner })
+            } catch (err) {
+                assert.equal(err.receipt.status, 0, "It should have thrown")
+                return
+            }
+            assert.isFalse(true, "It should have thrown")
+        })
+
+        /* TODO
+        it('changes support threshold thru voting', async () => {
```
izqui: Just a note so we don't forget the todo :)

kits/democracy/contracts/DemocracyKit.sol
bingen: Oh, yes, you are right, thanks! What do you think it's the best way to accomplish that? Using something else different than `NO_ENTITY` and `ANY_ENTITY` like for instance `address(1)`? Is there a better way? Maybe we could add a function in `ACL.sol` called `burnPermission` to make it easier and more clear.

izqui: I think the least confusing is going to be adding it to the ACL. Maybe rather than `burnPermission` we can call it `lockPermission`, `freezePermission` or `burnPermissionManager` as if someone already had the permission they will be able to continue performing the actions forever, as there will never be a manager again.

Actually `petrifyPermission` could be a good name as well 😅 Paging @sohkai for his input on this too.

kits/beta-base/contracts/BetaKitBase.sol
```diff
@@ -67,13 +63,18 @@ contract BetaTemplateBase {
...
+        emit InstalledApp(voting, appIds[uint8(Apps.Voting)]);
         Vault vault = Vault(dao.newAppInstance(appIds[uint8(Apps.Vault)], latestVersionAppBase(appIds[uint8(Apps.Vault)])));
-        InstalledApp(vault, appIds[uint8(Apps.Vault)]);
+        emit InstalledApp(vault, appIds[uint8(Apps.Vault)]);
         Finance finance = Finance(dao.newAppInstance(appIds[uint8(Apps.Finance)], latestVersionAppBase(appIds[uint8(Apps.Finance)])));
-        InstalledApp(finance, appIds[uint8(Apps.Finance)]);
-        TokenManager tokenManager = TokenManager(dao.newAppInstance(appIds[uint8(Apps.TokenManager)], latestVersionAppBase(appIds[uint8(Apps.TokenManager)])));
-        InstalledApp(tokenManager, appIds[uint8(Apps.TokenManager)]);
+        emit InstalledApp(finance, appIds[uint8(Apps.Finance)]);
+        TokenManager tokenManager = TokenManager(
```
izqui: Why are we only splitting into multiple lines in the case of the token manager?

kits/index_local.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
```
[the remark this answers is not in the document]
bingen: Nope, actually it's in `.gitignore` for kits that use it. I'll remove it.

kits/multisig/contracts/MultisigKit.sol
```diff
@@ -28,9 +28,11 @@ contract MultisigKit is BetaKitBase {
     }
 
     function newInstance(string name, address[] signers, uint256 neededSignatures) external {
-        require(signers.length > 0);
+        require(signers.length > 0 && neededSignatures > 0);
+        require(neededSignatures <= signers.length);
+        // We can avoid safemath checks here as it's very unlikely a user will pass in enough
+        // signers to cause this to overflow
         uint256 neededSignaturesE18 = neededSignatures * 10 ** 18;
-        require(neededSignaturesE18 >= signers.length);
```
bingen: It's probably nicer this way, but not so sure this is a fix:
- `neededSignatures > 0` was implicitly checked with `neededSignaturesE18 >= signers.length`
- `neededSignatures <= signers.length` will be checked on Voting app initialization, unless `signers.length` is bigger than `10^18`, which is assumed not to happen (and I guess that's why we are removing `neededSignaturesE18 >= signers.length`)
- I would remove `neededSignaturesE18` variable, or at least move it below to the only place it is used.

sohkai: > neededSignatures <= signers.length will be checked on Voting app initialization,

But the user will lose almost 90% (made up, but probably in that range) of their gas if we wait until then to do this check.

> I would remove neededSignaturesE18 variable, or at least move it below to the only place it is used.

Sounds good!

The reason I originally thought this was a bug was because `neededSignaturesE18 >= signers.length` looks like an unreasonable check (any `neededSignatures > 1` would pass).

bingen: > But the user will lose almost 90% (made up, but probably in that range) of their gas if we wait until then to do this check.
Good point. Let's leave it then. Although I would always prioritize low gas consumption for successful executions than for failures.
> any neededSignatures > 0 would pass
Well, yes, as long as `signers.length` is lower than 10^18, which is a good assumption, haha.

kits/democracy/contracts/DemocracyKit.sol
```diff
@@ -0,0 +1,63 @@
...
+    )
+        external
+    {
+        MiniMeToken token = popTokenCache(msg.sender);
+        Voting voting = createDAO(
+            name,
+            token,
+            holders,
+            tokens,
+            uint256(-1)
```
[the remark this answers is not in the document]
bingen: But if we use `0` then the token won't be transferable, right?
https://github.com/aragon/dao-kits/blob/aragonos-4/kits/beta-base/contracts/BetaKitBase.sol#L95
Do you want me to change that condition too?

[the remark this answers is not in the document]
sohkai: Ohhh good point, didn't realize it was used there. Let's keep it this way 👍.

kits/beta-base/contracts/BetaKitBase.sol
```diff
@@ -61,16 +65,42 @@ contract BetaKitBase is KitBase {
 
         acl.createPermission(this, dao, dao.APP_MANAGER_ROLE(), this);
 
-        Voting voting = Voting(dao.newAppInstance(appIds[uint8(Apps.Voting)], latestVersionAppBase(appIds[uint8(Apps.Voting)])));
+        Voting voting = Voting(
+            dao.newAppInstance(
+                appIds[uint8(Apps.Voting)],
+                latestVersionAppBase(appIds[uint8(Apps.Voting)]),
+                new bytes(0),
+                true
```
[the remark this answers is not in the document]
bingen: Do you mean the doing an extra `setApp`? Otherwise it's hard to retrieve apps from Kernel. Besides, I see at least `Voting` and `TokenManager` playing a "default" role in Democracy and Multisig DAOs.

kits/democracy/contracts/DemocracyKit.sol
```diff
@@ -38,13 +38,15 @@ contract DemocracyKit is BetaKitBase {
...
-        Voting voting = createDAO(
+        Kernel dao = createDAO(
             name,
             token,
             holders,
             tokens,
             uint256(-1)
         );
+
+        Voting voting = Voting(dao.getApp(dao.APP_ADDR_NAMESPACE(), appIds[uint8(Apps.Voting)]));
```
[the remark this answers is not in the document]
bingen: Oh, I guess this answers my previous comment too, haha. I think I misunderstood your initial comment. I thought that the point of it was to just get the DAO from `createDao`. Keep in mind that we would need to return `TokenManager` too, see [here](https://github.com/aragon/dao-kits/blob/issue2/kits/multisig/contracts/MultisigKit.sol#L69)

kits/beta-base/contracts/BetaKitBase.sol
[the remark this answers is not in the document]
izqui: I agree with @sohkai here, Democracy and Multisig DAOs can also be extended by installing extra instances of those apps. Also there is really no advantage in having them as default apps.

kits/beta-base/contracts/BetaKitBase.sol
```diff
@@ -1,24 +1,21 @@
-pragma solidity 0.4.18;
```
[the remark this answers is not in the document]
izqui: They serve no purpose at the moment, but it will make sense once kits stop being hardcoded in the app. We can add them for consistency or remove it from bare-kit for the time being.

Also just checked the CLI and there should be no problem in publishing a package without a manifest (actually we should probably add a warning).

kits/beta-base/contracts/BetaKitBase.sol
```diff
@@ -0,0 +1,160 @@
...
+        acl.createPermission(finance, vault, vault.TRANSFER_ROLE(), voting);
+        acl.createPermission(voting, finance, finance.CREATE_PAYMENTS_ROLE(), voting);
+        acl.createPermission(voting, finance, finance.EXECUTE_PAYMENTS_ROLE(), voting);
+        acl.createPermission(voting, finance, finance.DISABLE_PAYMENTS_ROLE(), voting);
+        acl.createPermission(voting, tokenManager, tokenManager.ASSIGN_ROLE(), voting);
+        acl.createPermission(voting, tokenManager, tokenManager.REVOKE_VESTINGS_ROLE(), voting);
+
+        // App inits
+        vault.initialize();
+        finance.initialize(vault, uint64(-1) - uint64(now)); // yuge period
```
izqui: I think we should set a reasonable period duration like 1 month, otherwise no DAO will be able to start using accounting periods in the future as the first period will never end and [changing period duration only affects the next period](https://github.com/aragon/aragon-apps/blob/master/apps/finance/contracts/Finance.sol#L550).

kits/beta-base/contracts/BetaKitBase.sol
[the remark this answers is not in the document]
bingen: Ok, reverting it then, although I still kind of disagree ;-)
- Having extra instances doesn't mean you can't have a default. Generally speaking, it's actually the opposite: IMO default makes sense precisely when you have several options (in the sense that when you have just one option you don't need to define a default). In this particular case, I think that a "Democracy DAO" having a default Voting app is a very natural thing. (Same applies for Multisig)
- The advantage is making it easy to access them with `dao.getApp`, plus the reasoning above.

kits/beta-base/contracts/BetaKitBase.sol
```diff
@@ -0,0 +1,160 @@
...
+        internal
+        returns (Kernel)
+    {
+        require(holders.length == stakes.length);
+
+        Kernel dao = fac.newDAO(this);
+
+        ACL acl = ACL(dao.acl());
+
+        acl.createPermission(this, dao, dao.APP_MANAGER_ROLE(), this);
```
izqui: Can save `dao.APP_MANAGER_ROLE()` and some other roles that are used more than once as a local variable so we don't do an extra call? If it triggers `stack too deep error` we can leave it for later.

kits/democracy/contracts/DemocracyKit.sol
```diff
@@ -0,0 +1,63 @@
...
+        voting.initialize(
+            token,
+            supportNeeded,
+            minAcceptanceQuorum,
+            voteDuration
+        );
+
+        // Burn support modification permission
+        ACL acl = ACL(dao.acl());
```
izqui: If we return return the Voting app as @sohkai proposes, maybe we can also return the `ACL` so we save a call.

bingen: But if we return `ACL`, `Voting`, and `TokenManager`, what do we want `Kernel` for? I'm afraid I'm confused.

izqui: In the case `Democracy` and `Multisig` they don't need a kernel reference as all the apps have already been deployed (but other future kits that use `beta-base` might), but as @sohkai said it feels weird that `createDAO` doesn't return a reference to the main DAO contract.

I think for consistency we could return the Kernel, ACL and the 4 apps that were deployed, even if our kits only need some of them.

sohkai: Originally I was thinking about the shotgun approach and just returning everything but didn't want to bloat the return size too much.

But if the alternative is to need calls to get the same info anyway, then we might as well return some extra data.

bingen: Returning more than 3 objects was throwing the following error: `CompilerError: Stack too deep, try removing local variables.`

PR #27, opened 2018-09-25
MultisigTemplate: only TokenManager should be allowed to create new votes

### LINKED ISSUE #2: MultisigTemplate: only TokenManager should be allowed to create new votes
The expected behaviour when dealing with a multisig is that only signers are allowed to propose new actions. BetaTemplateBase has the same permission for creating votes:

```solidity
acl.createPermission(ANY_ENTITY, voting, voting.CREATE_VOTES_ROLE(), voting);
```

bingen:
Closes #2.


bingen:
:+1:

kits/democracy/readme.md
```diff
@@ -29,15 +29,15 @@ npm run deploy:rinkeby
 
 | App           | Permission       | Grantee | Manager |
 |---------------|------------------|---------|---------|
+| Voting        | CREATE_VOTES     | Any     | Voting  |
 | Voting        | MODIFY_QUORUM    | Voting  | Voting  |
+| Voting        | MODIFY_SUPPORT   | None    | None    |
```
sohkai: Should this be "burned" for the manager?

kits/multisig/contracts/MultisigKit.sol
```diff
@@ -59,8 +61,12 @@ contract MultisigKit is BetaKitBase {
             1825 days // ~5 years
         );
 
-        // support modification permission
         ACL acl = ACL(Kernel(voting.kernel()).acl());
+
+        // create vote permission
+        acl.createPermission(tokenManager, voting, voting.CREATE_VOTES_ROLE(), voting);
```
izqui: Just a note: We need to modify the radspec string for the voting forward function, as when the token manager forwards to the voting app the vote is not casted (as discussed offline)

2018-09-27 at 9 53 18 pm

bingen: Yes, but how should we change it? the thing is that vote is going to be cast if `msg.sender` is a token holder, in this case sender is Token Manager, so that doesn't hold (and even if it did, it would be confusing as it wouldn't be the original sender's vote), but how do we evaluate that in a radspec string?

sohkai: For now I think we can change the text to say "casts a support vote if possible" and in the user documentation state that it's not possible in the event the create vote is forwarded by someone else.

Later on, we can add radspec helpers to detect this scenario (or improve this flow altogether somehow).

See https://github.com/aragon/aragon-apps/pull/470.

kits/democracy/readme.md
```diff
@@ -25,6 +25,20 @@ Then just:
...
+|---------------|------------------|---------|---------|
+| Voting        | CREATE_VOTES     | Any     | Voting  |
+| Voting        | MODIFY_QUORUM    | Voting  | Voting  |
+| Voting        | MODIFY_SUPPORT   | None    | Burned  |
+| Vault         | TRANSFER         | Finance | Voting  |
+| Finance       | CREATE_PAYMENTS  | Voting  | Voting  |
+| Finance       | EXECUTE_PAYMENTS | Voting  | Voting  |
+| Finance       | DISABLE_PAYMENTS | Voting  | Voting  |
+| Token Manager | ASSIGN           | Voting  | Voting  |
+| Token Manager | REVOKE_VESTINGS  | Voting  | Voting  |
```
sohkai: Ah, actually, it'd be nice to include the roles of the `ACL`, `Kernel`, and `EVMScriptRegistry` here too :).

PR #32, opened 2018-10-09
Survey: check there's no EVM Registry support enabled

### LINKED ISSUE #29: Add check to Survey Kit that the DAOFactory does not have EVMScript support enabled
The comment in https://github.com/aragon/dao-kits/blob/master/kits/survey/contracts/SurveyKit.sol#L27 is a little bit precarious.

Hopefully should be as simple as `require(_fac.regFactory == address(0))`...

bingen:
Closes #29.

### DIFF
--- a/kits/survey/contracts/SurveyKit.sol
+++ b/kits/survey/contracts/SurveyKit.sol
@@ -24,7 +24,7 @@ contract SurveyKit is APMNamehash, KitBase {
 
     constructor(DAOFactory _fac, ENS _ens) KitBase(_fac, _ens) public {
         // factory must be set up w/o EVMScript support
-        //require(address(_fac.regFactory()) == address(0));
+        require(address(_fac.regFactory()) == address(0));
     }
 
     function newInstance(
--- a/kits/survey/contracts/SurveyKit.sol
+++ b/kits/survey/contracts/SurveyKit.sol
@@ -22,8 +22,10 @@ contract SurveyKit is APMNamehash, KitBase {
 
     event DeployInstance(address dao, address indexed token);
 
-    // factory must be set up w/o EVMScript support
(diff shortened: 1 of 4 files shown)


kits/survey/contracts/SurveyKit.sol
```diff
@@ -22,8 +22,10 @@ contract SurveyKit is APMNamehash, KitBase {
 
     event DeployInstance(address dao, address indexed token);
 
-    // factory must be set up w/o EVMScript support
-    constructor(DAOFactory _fac, ENS _ens) KitBase(_fac, _ens) public {}
+    constructor(DAOFactory _fac, ENS _ens) KitBase(_fac, _ens) public {
+        // factory must be set up w/o EVMScript support
+        //require(address(_fac.regFactory()) == address(0));
```
sohkai: Should this be uncommented?

bingen: Oops!! :open_mouth: I was testing if it was working without it and forgot to put it back. Thanks!!

kits/survey/scripts/deploy.js
```diff
@@ -15,26 +16,39 @@ const newRepo = async (apm, name, acc, contract, contentURI = "ipfs:") => {
...
+
+  console.log('ENS', ens.address)
+  console.log('Owner', owner)
 
   const apmAddr = await artifacts.require('PublicResolver').at(await ens.resolver(namehash('aragonpm.eth'))).addr(namehash('aragonpm.eth'))
+  const apm = artifacts.require('APMRegistry').at(apmAddr)
 
+  // get network
+  if (process.argv.length < 5) {
+    errorOut('Usage: truffle exec --network <network> scripts/deploy.js')
```
sohkai: We should do this right away at the start to avoid spurious console output.

PR #33, opened 2018-10-11
[WIP] AGP23 kit

izqui:
Currently using the kit uses a bit more than 8M gas so it cannot be used in Rinkeby without modifications, we can split the token creation process to another transaction as we are doing right now in Rinkeby.

See [readme](https://github.com/aragon/dao-kits/blob/agp23-kit/kits/agp23/readme.md) for the permissions table and instructions for usage. I have been able to run it on a local chain with a modified unreleased version of Aragen.


sohkai:
Closing as [AGP-1](https://github.com/aragon/dao-kits/pull/47) supersedes AGP-23.

PR #35, opened 2018-10-19
 Prepare democracy and multisig to be published with aragon-cli

izqui:
This makes production kits ready to be published with the Aragon CLI.

I have made changed to aragen as well, that can now deploy these two kits: https://github.com/aragon/aragen/pull/15/commits/296b310f6e3cdab5c8625922bfdb4335e4df287c

### DIFF
--- a/kits/beta-base/scripts/deploy_kit.js
+++ b/kits/beta-base/scripts/deploy_kit.js
@@ -13,6 +13,8 @@ const appIds = apps.map(app => namehash(require(`@aragon/apps-${app}/arapp`).app
 const globalArtifacts = this.artifacts // Not injected unless called directly via truffle
 const defaultOwner = process.env.OWNER
 const defaultENSAddress = process.env.ENS
+const defaultDAOFactoryAddress = process.env.DAO_FACTORY
+const defaultMinimeTokenFactoryAddress = process.env.MINIME_TOKEN_FACTORY
 
 module.exports = async (
   truffleExecCallback,
@@ -20,9 +22,12 @@ module.exports = async (
     artifacts = globalArtifacts,
     owner = defaultOwner,
     ensAddress = defaultENSAddress,
+    daoFactoryAddress = defaultDAOFactoryAddress,
+    minimeTokenFactoryAddress = defaultMinimeTokenFactoryAddress,
     kitName,
     network,
-    verbose = true
+    verbose = true,
+    returnKit = false
   } = {}
 ) => {
   const log = (...args) => {
@@ -32,6 +37,7 @@ module.exports = async (
   log(`${kitName} in ${network} network`)
 
   const MiniMeTokenFactory = artifacts.require('MiniMeTokenFactory')
+  const DAOFactory = artifacts.require('DAOFactory')
   const ENS = artifacts.require('ENS')
 
   const newRepo = async (apm, name, acc, contract) => {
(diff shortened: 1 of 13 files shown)


kits/multisig/package.json
```diff
@@ -18,7 +18,9 @@
     "coverage": "./node_modules/@aragon/test-helpers/run-coverage.sh",
     "console": "truffle dev",
     "ganache-cli:dev": "./node_modules/@aragon/test-helpers/ganache-cli.sh",
-    "ganache-cli:coverage": "SOLIDITY_COVERAGE=true npm run ganache-cli:dev"
+    "ganache-cli:coverage": "SOLIDITY_COVERAGE=true npm run ganache-cli:dev",
+    "publish:rpc": "aragon apm publish $(npm run deploy:kit:rpc | tail -n 1) --network rpc",
+    "deploy:kit:rpc": "aragon contracts compile --network rpc && aragon contracts exec --network rpc scripts/multisig-deploy.js"
```
sohkai: It seems really weird we need to use `--network rpc` for `aragon contracts compile` but I'm guessing the CLI requires it?

kits/multisig/contracts/MultisigKit.sol
```diff
@@ -58,8 +75,8 @@ contract MultisigKit is BetaKitBase {
         uint256 multisigSupport = neededSignatures * 10 ** 18 / signers.length - 1;
         voting.initialize(
             token,
-            multisigSupport,
-            multisigSupport,
+            uint64(multisigSupport),
+            uint64(multisigSupport),
```
sohkai: Could we use the [`Uint256Helpers`](https://github.com/aragon/aragonOS/blob/dev/contracts/common/Uint256Helpers.sol) here and in the democracy kit?

kits/multisig/contracts/MultisigKit.sol
```diff
@@ -14,7 +14,24 @@ contract MultisigKit is BetaKitBase {
         BetaKitBase(_fac, _ens, _minimeFac, _aragonID, _appIds) public
     {}
 
-    function newToken(string name, string symbol) external returns (MiniMeToken token) {
+    function newDemoInstance(string name) external {
+        address[] memory holders = new address[](1);
```
bingen: A Multisig 1 out of 1 is kind of funny ;-)

kits/beta-base/scripts/deploy_kit.js
```diff
@@ -40,36 +46,52 @@ module.exports = async (
...
   const kit = await artifacts.require(kitName).new(daoFactory.address, ens.address, minimeFac.address, aragonid, appIds)
 
+  if (returnKit) {
+    return kit
+  }
+
   const ts = [ { name: kitName, address: kit.address } ]
   log(ts)
 
+  log('Creating APM package with owner', owner)
```
sohkai: I think this only happens if the network is `devnet`?

kits/democracy/contracts/DemocracyKit.sol
```diff
@@ -15,7 +15,44 @@ contract DemocracyKit is BetaKitBase {
         public
     {}
 
-    function newToken(string name, string symbol) external returns (MiniMeToken token) {
+    function newDemoInstance(string name) external {
```
sohkai: Not a super big fan of having this available by default; this seems like it'd be better as a "demo config" that's specified and parsed off-chain by the CLI.

kits/beta-base/scripts/deploy_kit.js
```diff
@@ -40,36 +46,52 @@ module.exports = async (
...
-    log("Local testing network, exiting...")
-    return;
-  }
-
   let indexFileName
-  if (network != 'rpc' && network != 'devnet') {
-    indexFileName = 'index.js'
-  } else {
-    indexFileName = 'index_local.js'
+  if (!returnKit) {
```
bingen: If `returnKit` is false and no ENS address is provided by param or in `ENV`, `indexFileName` below in line 58 will be undefined.

kits/democracy/scripts/democracy-deploy.js
```diff
@@ -0,0 +1,14 @@
+const deployKit = require('@aragon/kits-beta-base/scripts/deploy_kit.js')
+
+module.exports = async (callback) => {
+  const deployConfig = {
```
bingen: Just curious, where are you defining `ENS` for `deployKit` when you run `deploy:kit:rpc`? `ensAddress` nor `network` are being passed, so I understand it should be in `process.ENV`, but I don't see where it's set.

kits/multisig/scripts/multisig-deploy.js
```diff
@@ -0,0 +1,14 @@
+const deployKit = require('@aragon/kits-beta-base/scripts/deploy_kit.js')
+
+module.exports = async (callback) => {
+  const deployConfig = {
```
bingen: Same question as with democracy.

kits/multisig/package.json
izqui: Yup, agreed. It is a bug in the CLI which requires passing a network for that command.

izqui: Will be fixed here: https://github.com/aragon/aragon-cli/pull/230/files#r227244994

kits/democracy/contracts/DemocracyKit.sol
izqui: Good point, let's remove it for now and figure out how to do this well in the future.

kits/beta-base/scripts/deploy_kit.js
izqui: That's true, adding a explicit error.

kits/democracy/scripts/democracy-deploy.js
izqui: Yes, it assumes `ENS` has been passed as an environment variable. In the case of aragen it gets exported here: https://github.com/aragon/aragen/blob/863cd750d8d5c326c624ebee13362d72b316b406/scripts/set-env

izqui:
Address most comments, waiting for the release candidate versions to be cut to update dependencies here, and test again.

PR #38, opened 2018-10-24
Add deploy logger to production kits deployment script

izqui:


kits/beta-base/scripts/deploy_kit.js
```diff
@@ -5,6 +5,7 @@ const fs = require('fs')
 const namehash = require('eth-ens-namehash').hash
 
 const deployDAOFactory = require('@aragon/os/scripts/deploy-daofactory.js')
+const logDeploy = require('@aragon/os/scripts/helpers/deploy-logger')
```
sohkai: I'm getting the sense this should've been made its own package.

izqui: Yep, we can also move aragonOS' `truffle-config` to that same package as it also gets imported everywhere.

PR #40, opened 2018-10-24
Allow to test in several networks

bingen:
Needs:

https://github.com/aragon/aragonOS/pull/447
https://github.com/aragon/geth-devnet/pull/2


helpers/networks.js
```diff
@@ -0,0 +1,22 @@
+const getNetworkNameFromId= (networks, id) => {
```
izqui:
```suggestion
const getNetworkNameFromId = (networks, id) => {
```

helpers/networks.js
```diff
@@ -0,0 +1,22 @@
...
+  const defaultNetwork = 'devnet'
+  for (let n in networks) {
+    if (networks[n].network_id == id) {
+      return n
+    }
+  }
+  return defaultNetwork
+}
+const getNetworkId = () =>
+        new Promise(
```
izqui: Indent :)

kits/democracy/index.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+  "networks": {
+    "rinkeby": {
+      "ens": "0xfe03625ea880a8cba336f9b5ad6e15b0a3b5a939",
+      "kits": [
+        {
+          "name": "DemocracyKit",
+          "address": "0x7501fF13969d4084223eDb26CC04FE385e6F860a"
```
izqui: Can we get them from ENS rather than hardcoding?

helpers/networks.js
bingen: My editor did that automatically, how should I indent it? Just 2 spaces from the beginning?

kits/beta-base/scripts/deploy_kit.js
```diff
@@ -57,16 +57,16 @@ module.exports = async (
     } else {
       arappFileName = 'arapp_local.json'
     }
+    if (!ensAddress) {
+      const betaArapp = require('../' + arappFileName)
+      ensAddress = betaArapp.environments[network].registry
+    }
+  } else {
+    if (!ensAddress) {
```
sohkai: You can just do an `else if {` :)

bingen: Haha, I was actually expecting this. I did this way first, then I changed to `else if`, but then I moved back to how it is now, because it looked way better to me, because of the clear separation of the 2 use cases of the script (returning kit or not) and because of the parallelism with the previous `if` block. But if you like it better with an `else if` I can change it, of course ;-)
(I don't know what you mean by "if this should be an if block by itself")

sohkai: > (I don't know what you mean by "if this should be an if block by itself")

Just always checking if there's an `ensAddress`, since it seems like an invariant for this script to run. But we can leave it like this :).

bingen: Oh, yes, now I get it, so you mean checking that even in when `returnKit` is `false`, leave that check outside so we also check that it was able to get it from `arapp.json` if it didn't have it before. Yes, that makes total sense, I'll change it.

PR #46, opened 2018-11-08
Split KitBase out of Bare into its own package

bingen:


bingen:
Oh, good catch. I ran tests for the ones that have them and did "greps" searching for "kits-bare", but missed this obvious one, :-o. Fixing it now.

PR #47, opened 2018-11-09
Add AGP-1 Kit

bingen:
TODO:

- [X] Add tests
- [x] Deploy in Rinkeby


kits/agp1/contracts/AGP1Kit.sol
```diff
@@ -0,0 +1,86 @@
...
+import "@aragon/kits-base/contracts/KitBase.sol";
+
+
+contract AGP1Kit is KitBase, APMNamehash, IsContract {
+
+    uint64 constant public MAIN_VOTING_SUPPORT = 50 * 10**16; // > 50%
+    uint64 constant public MAIN_VOTING_QUORUM = 0; // Just 1 vote is enough
+    uint64 constant public MAIN_VOTING_VOTE_TIME = 48 hours;
+
+    uint64 constant public META_TRACK_VOTING_SUPPORT = 666666666666666667; // > two thirds
```
izqui: Tiny detail: Ending in a 6 is the strict way of requiring > 2/3, as `666666666666666667` is already > 2/3 and the voting app has strict thresholds.

kits/agp1/contracts/AGP1Kit.sol
```diff
@@ -0,0 +1,86 @@
...
+    uint64 constant public META_TRACK_VOTING_QUORUM = 0; // Just 1 vote is enough
+    uint64 constant public META_TRACK_VOTING_VOTE_TIME = 48 hours;
+
+    uint64 constant public FINANCE_PERIOD_DURATION = 90 days;
```
izqui: I think I would make it `365.25 days / 4` so in the long run there are 4 periods per year. Alternatively we can make it `365 days / 4` which would be less weird.

kits/agp1/scripts/deploy_agp1.js
```diff
@@ -0,0 +1,114 @@
...
+  console.log(await ens.owner(appIds[0]))
+  const apm = artifacts.require('APMRegistry').at(apmAddress)
+
+  for (let i = 0; i < apps.length; i++) {
+    if (await ens.owner(appIds[i]) == '0x0000000000000000000000000000000000000000') {
+      errorOut(`Missing app ${apps[i]}, aborting.`)
+    }
+  }
+
+  const agp1Kit = await artifacts.require(kitName).new(daoFactoryAddress, ensAddress)
```
izqui: We should use the `deploy-logger` from aragonOS here

kits/agp1/test/agp1.js
```diff
@@ -0,0 +1,256 @@
...
+    const holder16 = accounts[1]
+    const holder33 = accounts[2]
+    const holder51 = accounts[3]
+    const nonHolder = accounts[4]
+
+    before(async () => {
+        const networkName = (await getNetwork(networks)).name
+        if (networkName == 'devnet' || networkName == 'rpc') {
+            // transfer some ETH to other accounts
+            await web3.eth.sendTransaction({ from: owner, to: holder16, value: web3.toWei(1, 'ether') })
```
izqui: Is this only needed for when we are testing using geth? AFAIK, ganache funds the first 10 addresses with 100 eth.

kits/agp1/test/agp1.js
```diff
@@ -0,0 +1,256 @@
...
+            financeAddress,
+            vaultAddress,
+            votingAddress,
+            metaTrackVotingAddress
+        } = await deployAgp1(null, {artifacts, web3, owner})
+
+        daoAddress = agp1Address
+        tokenAddress = minimeTokenAddress
+
+        finance = await Finance.at(financeAddress)
```
izqui: I dont think these two awaits are needed

kits/agp1/test/agp1.js
```diff
@@ -0,0 +1,256 @@
...
+            // vault
+            await checkRole(vault.address, await vault.TRANSFER_ROLE(), metaTrackVoting.address, 'Vault', 'TRANSFER', finance.address)
+
+            // finance
+            await checkRole(finance.address, await finance.CREATE_PAYMENTS_ROLE(), metaTrackVoting.address, 'Finance', 'CREATE_PAYMENTS', voting.address)
+            await checkRole(finance.address, await finance.EXECUTE_PAYMENTS_ROLE(), metaTrackVoting.address, 'Finance', 'EXECUTE_PAYMENTS', voting.address)
+            await checkRole(finance.address, await finance.MANAGE_PAYMENTS_ROLE(), metaTrackVoting.address, 'Finance', 'MANAGE_PAYMENTS', voting.address)
+        })
+
+        context('creating vote', () => {
```
izqui: Can we test this for both voting apps using a generative test?

izqui: I would also add a test for how the metaTrackVoting app can execute meta actions, but it would revert if the voting app tries to perform those actions.

kits/agp1/test/agp1.js
bingen: Yes, I used this to test it with our docker image, but I think it's enough to test with ganache, so I'll remove it.

izqui: I would maintain the real docker tests if possible actually

bingen: Ok, but I'll put it in a separate file because `assertRevert` and `timeTravel` were not working there for me.

kits/agp1/contracts/AGP1Kit.sol
```diff
@@ -0,0 +1,86 @@
...
+
+    uint64 constant public META_TRACK_VOTING_SUPPORT = 666666666666666666; // > two thirds
+    uint64 constant public META_TRACK_VOTING_QUORUM = 0; // Just 1 vote is enough
+    uint64 constant public META_TRACK_VOTING_VOTE_TIME = 48 hours;
+
+    uint64 constant public FINANCE_PERIOD_DURATION = 7889400; // 365.25 days / 4
+
+    bytes32 constant public financeAppId = apmNamehash("finance");
+    bytes32 constant public vaultAppId = apmNamehash("vault");
+    bytes32 constant public votingAppId = apmNamehash("voting");
```
sohkai: We don't really _need_ any of these constants to be public, but it probably doesn't matter here.

kits/agp1/contracts/AGP1Kit.sol
```diff
@@ -0,0 +1,86 @@
...
+
+    constructor(DAOFactory _fac, ENS _ens) KitBase(_fac, _ens) public {
+        require(isContract(address(_fac.regFactory())));
+    }
+
+    function newInstance(MiniMeToken ant, address multisig) external returns (Kernel) {
```
sohkai: Just style, but we should keep to the `_` prefix here for arguments.

kits/agp1/contracts/AGP1Kit.sol
```diff
@@ -0,0 +1,86 @@
...
+
+        Voting metaTrackVoting = Voting(dao.newAppInstance(votingAppId, latestVersionAppBase(votingAppId)));
+        emit InstalledApp(metaTrackVoting, votingAppId);
+
+        // permissions
+        acl.createPermission(multisig, voting, voting.CREATE_VOTES_ROLE(), multisig);
+        acl.createPermission(metaTrackVoting, voting, voting.MODIFY_QUORUM_ROLE(), metaTrackVoting);
+        acl.createPermission(metaTrackVoting, voting, voting.MODIFY_SUPPORT_ROLE(), metaTrackVoting);
+
+        acl.createPermission(multisig, metaTrackVoting, metaTrackVoting.CREATE_VOTES_ROLE(), multisig);
```
sohkai: Should the `CREATE_VOTES_ROLE` manager also be `metaTrackVoting`?

kits/agp1/scripts/deploy_agp1.js
```diff
@@ -0,0 +1,120 @@
...
+  } else {
+    log(`Using MiniMeToken: ${minimeTokenAddress}`)
+  }
+
+  const apmAddress = await artifacts.require('PublicResolver').at(await ens.resolver(namehash('aragonpm.eth'))).addr(namehash('aragonpm.eth'))
+  if (!apmAddress) {
+    errorOut('No APM found for ENS, aborting.')
+  }
+  log('APM', apmAddress);
+  console.log(await ens.owner(appIds[0]))
```
sohkai: Should this just use `log()`?

kits/agp1/scripts/deploy_agp1.js
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  log(`${kitName} with ENS ${ensAddress}, owner ${owner}`)
+
+  const TokenFactoryWrapper = artifacts.require('TokenFactoryWrapper')
+  const DAOFactory = artifacts.require('DAOFactory')
+  const ENS = artifacts.require('ENS')
+
+  if (!ensAddress) {
+    errorOut('ENS environment variable not passed, aborting.')
```
sohkai: I'm not sure, but are we missing an import or a definition for `errorOut()`? Seems like all the kits actually use this without a definition... is it injected globally by truffle?

kits/agp1/scripts/deploy_deps.js
```diff
@@ -0,0 +1,53 @@
...
+  let owner = process.env.OWNER
+  const accounts = await getAccounts(web3)
+  if (!owner) {
+    owner = accounts[0]
+    console.log('OWNER env variable not found, setting APM owner to the provider\'s first account')
+  }
+  console.log(`Deploying dependencies, owner: ${owner}`)
+
+  if (process.argv.length < 5) {
+    errorOut('Usage: truffle exec --network <network> scripts/deploy_deps.js')
```
sohkai: Same here for `errorOut()`

kits/agp1/contracts/AGP1Kit.sol
bingen: Hm, good question, not sure, but in our specs document is defined like this.

kits/agp1/scripts/deploy_agp1.js
bingen: Well, I just forgot it, I was using it at some point for debugging, but I think we can better remove it.

kits/agp1/contracts/AGP1Kit.sol
izqui: I don't think so, as the AA multisig may decide to deprecate the wallet and move to another one (which IMO shouldn't require token holder approval), or transfer the power over proposing votes to a separate mechanism.

kits/agp1/scripts/deploy_agp1.js
izqui: I have no idea, the first time I saw it was when bingen added it to the kit's code.

bingen: :astonished: haha, sorry, I'll remove it then

kits/agp1/contracts/AGP1Kit.sol
izqui: For kits i like appIds to be public as it helps sanity check and debugging

PR #48, opened 2018-11-12
Add ANT contracts and deployment script for testing

bingen:


bingen:
Yes, sounds good, I think so too, so closing it.

sohkai:
@bingen Maybe we should document the deployed ANT instance on Rinkeby in the `aragon-network-token` repo?

PR #49, opened 2018-11-14
 Enable CI for AGP-1 kit tests

izqui:


sohkai:
I think we might as well let the CI run it, but I've added configuration so that it doesn't block a run from passing.

PR #50, opened 2018-12-01
Restrict democracy create vote

### LINKED ISSUE #53: Change default permissions so that only an account who belongs to an organisation can create new votes.
Anyone who knows the URL of my organisation (it's blockchain so it's probably publicly visible) can create new votes.

These are the default settings:

image

I suggest only an account who owns a token in the organisation can create votes.

----

How did I discovered this issue?

I was testing the app and wanted to onboard some friends. Gave them instructions:
> Visit: https://rinkeby.aragon.org
> Open an existing organisation: foobar

I was planning to ask them for ETH address so that I could assign them tokens.

Realised that they can create an assignment themselves.

The process feels weird.

**Real life analogy:**
Onboarding at a company, it's the admin who creates mars@aragon.one

Or is it me filling the form to HR?

What if HR is publicly open and anyone can submit the form?

bingen:
Allow only token holders to create votes.
Fixes #53

### DIFF
--- a/kits/democracy/test/democracy.js
+++ b/kits/democracy/test/democracy.js
@@ -20,7 +20,13 @@ const getContract = name => artifacts.require(name)
 
 const pct16 = x => new web3.BigNumber(x).times(new web3.BigNumber(10).toPower(16))
 const getEventResult = (receipt, event, param) => receipt.logs.filter(l => l.event == event)[0].args[param]
-const createdVoteId = receipt => getEventResult(receipt, 'StartVote', 'voteId')
+const getVoteId = (receipt) => {
+    const logs = receipt.receipt.logs.filter(
+        l =>
+            l.topics[0] == web3.sha3('StartVote(uint256,address,string)')
+    )
+    return web3.toDecimal(logs[0].topics[1])
+}
 const getAppProxy = (receipt, id) => receipt.logs.filter(l => l.event == 'InstalledApp' && l.args.appId == id)[0].args.appProxy
 const networks = require("@aragon/os/truffle-config").networks
 const getNetwork = require('../../../helpers/networks.js')
@@ -149,7 +155,7 @@ contract('Democracy Kit', accounts => {
             await checkRole(reg.address, await reg.REGISTRY_MANAGER_ROLE(), votingAddress, 'EVMScriptRegistry', 'REGISTRY_MANAGER')
 
             // voting
-            await checkRole(votingAddress, await voting.CREATE_VOTES_ROLE(), votingAddress, 'Voting', 'CREATE_VOTES', await acl.ANY_ENTITY())
+            await checkRole(votingAddress, await voting.CREATE_VOTES_ROLE(), votingAddress, 'Voting', 'CREATE_VOTES', tokenManagerAddress)
             await checkRole(votingAddress, await voting.MODIFY_QUORUM_ROLE(), votingAddress, 'Voting', 'MODIFY_QUORUM')
             assert.equal(await acl.getPermissionManager(votingAddress, await voting.MODIFY_SUPPORT_ROLE()), await acl.BURN_ENTITY(), 'Voting MODIFY_SUPPORT Manager should be burned')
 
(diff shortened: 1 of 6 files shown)


kits/multisig/test/multisig.js
```diff
@@ -238,6 +238,32 @@ contract('Multisig Kit', accounts => {
...
+            it('throws when non-holder tries to create a vote directly', async () => {
+                const action = { to: executionTarget.address, calldata: executionTarget.contract.execute.getData() }
+                script = encodeCallScript([action])
+                try {
+                    await voting.newVote(script, 'metadata', { from: nonHolder })
+                } catch (err) {
+                    assert.equal(err.receipt.status, 0, "It should have thrown")
+                    return
+                }
+                assert.isFalse(true, "It should have thrown")
```
sohkai: Might be nice just to use the `assertRevert` helpers we use in `aragon-apps` (it's published as [`@aragon/test-helpers`](https://www.npmjs.com/package/@aragon/test-helpers))

kits/democracy/test/democracy.js
```diff
@@ -20,7 +20,13 @@ const getContract = name => artifacts.require(name)
 
 const pct16 = x => new web3.BigNumber(x).times(new web3.BigNumber(10).toPower(16))
 const getEventResult = (receipt, event, param) => receipt.logs.filter(l => l.event == event)[0].args[param]
-const createdVoteId = receipt => getEventResult(receipt, 'StartVote', 'voteId')
+const getVoteId = (receipt) => {
```
sohkai: Does `getEventResult()` not work for getting this? I was under the impressions truffle would parse the logs for us.

sohkai:
It looks like the democracy CI is failing due to low account balance issues, which definitely seems weird :/.

kits/multisig/test/multisig.js
bingen: If I recall correctly, that was not working with the geth node in our docker image.

kits/democracy/test/democracy.js
bingen: No, it was not working, because it's not the "first" call (now we are forwarding trough `TokenManager`), so logs don't show up in `receipt.logs.filter` and we have to look for them in `receipt.receipt.logs` instead.

bingen:
Yes, I don't know what's going on with CI, tests are passing locally for me. I'll try to figure it out.

kits/democracy/test/democracy.js
```diff
@@ -72,10 +72,10 @@ contract('Democracy Kit', accounts => {
...
         if (networkName == 'devnet' || networkName == 'rpc') {
             // transfer some ETH to other accounts
-            await web3.eth.sendTransaction({ from: owner, to: holder20, value: web3.toWei(1, 'ether') })
-            await web3.eth.sendTransaction({ from: owner, to: holder29, value: web3.toWei(1, 'ether') })
-            await web3.eth.sendTransaction({ from: owner, to: holder51, value: web3.toWei(1, 'ether') })
-            await web3.eth.sendTransaction({ from: owner, to: nonHolder, value: web3.toWei(1, 'ether') })
+            await web3.eth.sendTransaction({ from: owner, to: holder20, value: web3.toWei(10, 'ether') })
+            await web3.eth.sendTransaction({ from: owner, to: holder29, value: web3.toWei(10, 'ether') })
+            await web3.eth.sendTransaction({ from: owner, to: holder51, value: web3.toWei(10, 'ether') })
+            await web3.eth.sendTransaction({ from: owner, to: nonHolder, value: web3.toWei(10, 'ether') })
```
sohkai: 😄

bingen: Haha, yes, so much time wasted for such an easy solution. Anyway, I still think there's something weird going on here, because it was randomly failing and because if I'm not wrong the transaction which failed was casting a vote and the account performing it had 0.6 ETH (don't know what gas price we have at that node but it should be quite high).

izqui:
Given that we have now merged this into master and we will be redeploying kits this week, we would need to update the user doc to reflect this change in 0.6.2 cc @john-light

PR #56, opened 2018-12-12
chore: update license to Aragon Association

sohkai:


LICENSE
```diff
@@ -1,7 +1,9 @@
+                     Copyright (C) 2017-2018
+
                     GNU GENERAL PUBLIC LICENSE
                        Version 3, 29 June 2007
 
- Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
+ Copyright (C) Aragon Association <https://aragon.org/>
```
bingen: Are you sure we have to / can modify this?
See this: https://opensource.stackexchange.com/a/4024 (or two lines below)

sohkai: D: Good catch. I saw a few other projects changing this and assumed this was how you add your organization. I'll change them all again to be in the header and keep this line.

PR #58, opened 2018-12-12
feat(Beta): allow different name for token and organization

sohkai:
Adds the ability to specify a different token name from the organization's registered aragonID name.

This is especially important, as aragonID names lose a lot of freedom during UTS46 normalization (e.g. all uppercase characters).

Parameterizes the tests so that the suite is run for both types of organization creation, and adds a few more tests:

- 0d24ec6 and 4ce2bdb add the test parameterization
- e189bee adds some more tests

### DIFF
--- a/kits/multisig/test/multisig.js
+++ b/kits/multisig/test/multisig.js
@@ -1,4 +1,5 @@
 require('dotenv').config({ path: './node_modules/@aragon/kits-beta-base/.env'})
+
 const getBlockNumber = require('@aragon/test-helpers/blockNumber')(web3)
 const getBlock = require('@aragon/test-helpers/block')(web3)
 //const timeTravel = require('@aragon/test-helpers/timeTravel')(web3)
@@ -7,6 +8,11 @@ const namehash = require('eth-ens-namehash').hash
 const keccak256 = require('js-sha3').keccak_256
 
 const { encodeCallScript, EMPTY_SCRIPT } = require('@aragon/test-helpers/evmScript')
+
+const ENS = artifacts.require('ENS')
+const PublicResolver = artifacts.require('PublicResolver')
+
+const MiniMeToken = artifacts.require('MiniMeToken')
 
 const Finance = artifacts.require('Finance')
 const TokenManager = artifacts.require('TokenManager')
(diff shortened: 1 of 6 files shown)


kits/multisig/test/multisig.js
```diff
@@ -77,264 +77,285 @@ contract('Multisig Kit', accounts => {
...
-                try {
-                    await voting.newVote(script, 'metadata', { from: nonHolder })
-                } catch (err) {
-                    assert.equal(err.receipt.status, 0, "It should have thrown")
-                    return
+    // Test when organization is created in one call with `newTokenAndInstance()` and in
+    // two calls with `newToken()` and `newInstance()`
+    const creationStyles = ['single', 'separate']
+    for (const creationStyle of creationStyles) {
+        context(`Creation through ${creationStyle} transaction`, () => {
```
sohkai: @bingen I'm trying to add a parameterized test so the same tests we had before can be run for organizations created through `newTokenAndInstance()`.

kits/multisig/test/multisig.js
```diff
@@ -76,270 +84,309 @@ contract('Multisig Kit', accounts => {
...
-            await checkRole(votingAddress, await voting.MODIFY_SUPPORT_ROLE(), votingAddress, 'Voting', 'MODIFY_SUPPORT')
+    // Test when organization is created in one call with `newTokenAndInstance()` and in
+    // two calls with `newToken()` and `newInstance()`
+    const creationStyles = ['single', 'separate']
+    for (const creationStyle of creationStyles) {
+        context(`> Creation through ${creationStyle} transaction`, () => {
+            let aragonId, tokenName, tokenSymbol
+
+            before(async () => {
+                aragonId = 'multisigdao-' + Math.floor(Math.random() * 1000)
```
sohkai: You would not believe how long it took me to realize that `namehash('MultisigDao-<number>.aragonid.eth')` doesn't create the same hash as the one the contract makes (`sha3(namehash('aragonid.eth') + sha3('MultisigDao-<number>'))`), due to the [UTS46 normalization](https://github.com/danfinlay/eth-ens-namehash/blob/master/index.js#L26) 😄 .

This bit was previously also giving aragonID incorrect ENS nodes to register (without the `Math.floor()`, `Math.random() * 1000` gives something like `123.1234`).

izqui:
@bingen can you take a look at these tests since it was you who mostly wrote them?

bingen:
Sure, I'll do!

PR #60, opened 2019-01-03
chore(bare): scripts for publishing to live networks

### LINKED ISSUE #59: Publish bare kit

izqui:
Close #59

### DIFF
--- a/kits/bare/package.json
+++ b/kits/bare/package.json
@@ -6,7 +6,12 @@
   "license": "GPL-3.0-or-later",
   "scripts": {
     "publish:rpc": "aragon apm publish major $(npm run deploy:kit:rpc | tail -n 1) --environment default",
-    "deploy:kit:rpc": "aragon contracts compile --all && aragon contracts exec --network rpc scripts/deploy.js"
+    "deploy:kit:rpc": "aragon contracts compile --all && aragon contracts exec --network rpc scripts/deploy.js",
+    "publish:staging": "aragon apm publish major $(npm run deploy:kit:rinkeby | tail -n 1) --environment staging",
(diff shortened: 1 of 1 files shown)


kits/bare/package.json
```diff
@@ -6,7 +6,12 @@
...
-    "deploy:kit:rpc": "aragon contracts compile --all && aragon contracts exec --network rpc scripts/deploy.js"
+    "deploy:kit:rpc": "aragon contracts compile --all && aragon contracts exec --network rpc scripts/deploy.js",
+    "publish:staging": "aragon apm publish major $(npm run deploy:kit:rinkeby | tail -n 1) --environment staging",
+    "publish:rinkeby": "aragon apm publish major $(npm run deploy:kit:rinkeby | tail -n 1) --environment rinkeby",
+    "deploy:kit:rinkeby": "aragon contracts compile --all && aragon contracts exec --network rinkeby scripts/deploy.js",
+    "publish:mainnet": "aragon apm publish major $(npm run deploy:kit:mainnet | tail -n 1) --environment mainnet",
+    "deploy:kit:mainnet": "aragon contracts compile --all && aragon contracts exec --network mainnet scripts/deploy.js"
   },
   "dependencies": {
     "@aragon/os": "^4.0.0",
```
izqui: Should we bump this?

izqui:
@bingen that’s passed directly to truffle

bingen:
Thanks for the clarification @izqui. I don't see the point of wrapping truffle into aragon-cli, but if we do this, I think it would be nice to have consistency, use `--environment` and then from it get the network to be passed to truffle (of course, out of the scope of this PR).
Actually, truffle doesn't need network for `compile` command, right?

izqui:
That's a good idea! Not really, but we do need it for `truffle exec`

PR #63, opened 2019-02-16
SurveyKit: modernize

sohkai:
Also includes:

- Some contract fixes for the kit
- Installs a Vault instance as the escape hatch (we have to set a contract as the escape hatch)
- Demo launch script, for easy local testing

TODO:

- [ ] Update Survey dependency once republished, and then retest

### DIFF
--- a/kits/survey/contracts/SurveyKit.sol
+++ b/kits/survey/contracts/SurveyKit.sol
@@ -4,6 +4,7 @@ import "@aragon/os/contracts/apm/APMNamehash.sol";
 import "@aragon/os/contracts/apm/Repo.sol";
 import "@aragon/os/contracts/factory/DAOFactory.sol";
 import "@aragon/os/contracts/kernel/Kernel.sol";
+import "@aragon/os/contracts/kernel/KernelConstants.sol";
 import "@aragon/os/contracts/acl/ACL.sol";
 import "@aragon/apps-shared-minime/contracts/MiniMeToken.sol";
 import "@aragon/os/contracts/lib/ens/ENS.sol";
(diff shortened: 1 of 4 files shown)


kits/survey/contracts/SurveyKit.sol
```diff
@@ -43,16 +42,17 @@ contract SurveyKit is APMNamehash, KitBase {
...
+        Vault vault = Vault(dao.newAppInstance(KERNEL_DEFAULT_VAULT_APP_ID, latestVersionAppBase(KERNEL_DEFAULT_VAULT_APP_ID)));
+        vault.initialize();
+
         // Set survey manager as the entity that can create votes and change participation
         // surveyManager can then give this permission to other entities
         acl.createPermission(surveyManager, survey, survey.CREATE_SURVEYS_ROLE(), surveyManager);
         acl.createPermission(surveyManager, survey, survey.MODIFY_PARTICIPATION_ROLE(), surveyManager);
+        acl.createPermission(surveyManager, vault, vault.TRANSFER_ROLE(), surveyManager);
         acl.grantPermission(surveyManager, dao, dao.APP_MANAGER_ROLE());
         acl.setPermissionManager(surveyManager, dao, dao.APP_MANAGER_ROLE());
```
sohkai: For some reason, `cleanupDAOPermissions()` always fails for me... it keeps complaining about the role not having an manager set (`ACL_AUTH_NO_MANAGER`).

cc @izqui

bingen: Yes, there's a function [`cleanupPermission`](https://github.com/aragon/dao-kits/blob/master/kits/base/contracts/KitBase.sol#L37) for cleaning permissions individually, but it's better to remove that line and rely on the other one. Actually the line before (L56) can be removed, as the permission is also set in `cleanupDAOPermissions`.

facuspagnuolo:
Closing in favor of https://github.com/aragon/aragon-apps/pull/915

PR #65, opened 2019-04-08
test: fix test parameterization variables

sohkai:
Similar to https://github.com/aragon/aragonOS/pull/506


sohkai:
Actually not an issue in this repo, since we used `const` appropriately.

PR #77, opened 2019-06-06
Fix scripts error handling

facuspagnuolo:
As we've been saying before, we do need to refactor these deployment scripts, there's code duplicated and it is not working flawless. This PR does not solve that at all, just tweaking it to at least not swallow thrown errors. 

I'm not aware of how the CLI interacts with these scripts, or if it expects them to behave in certain way. Note that I'm changing the way the truffle exec callback is handled in the scripts.

### DIFF
--- a/kits/democracy/scripts/deploy.js
+++ b/kits/democracy/scripts/deploy.js
@@ -4,7 +4,12 @@ const deploy_apm = require('@aragon/os/scripts/deploy-apm.js')
 const deploy_id = require('@aragon/id/scripts/deploy-beta-aragonid.js')
 const deploy_kit = require('@aragon/kits-beta-base/scripts/deploy_kit.js')
 
-module.exports = async (callback) => {
+const errorOut = (msg) => {
+  console.error(msg)
+  throw new Error(msg)
+}
+
+async function deploy() {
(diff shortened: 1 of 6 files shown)


kits/democracy/scripts/deploy.js
```diff
@@ -22,5 +27,9 @@ module.exports = async (callback) => {
   // aragonID
   await deploy_id(null, { artifacts, web3, ensAddress: ens.address })
 
-  await deploy_kit(null, { artifacts, kitName: 'democracy-kit', kitContractName: 'DemocracyKit', network: network, ensAddress: ens.address })
+  await deploy_kit(null, { artifacts, kitName: 'democracy-akropolis', kitContractName: 'DemocracyKit', network: network, ensAddress: ens.address })
```
bingen: Just curious, why this change? Anyway, it's definitely nicer! ;-) (although maybe less intuitive, and I see mltisig one hasn't been changed)

facuspagnuolo: My bad! 🤦🏼‍♂️
I carried a change from another fork, please merge #81 to fix it 🙏🏻

PR #79, opened 2019-06-11
Chore: Remove ethereum-data dir

facuspagnuolo:
The dev geth node we use comes with its own data dir, there is no need to have maintain a separate one.


facuspagnuolo:
Nvm, I checked with @bingen and we are using it to avoid waiting too much time for the target gas limit

sohkai:
@facuspagnuolo Perhaps we can update the data-dir in the dev geth node so we don't have to wait for gas at all?

PR #82, opened 2019-06-16
Update arrap.json with wsRPC

0xGabi:
Update arapp.json to use the `wsRPC` introduced with https://github.com/aragon/aragon-cli/pull/270, having it will solve issues like this: https://github.com/aragon/aragon-cli/issues/522.

### DIFF
--- a/kits/bare/arapp.json
+++ b/kits/bare/arapp.json
@@ -6,22 +6,46 @@
       "network": "rpc"
     },
     "rinkeby": {
+      "apm": {
+        "ipfs": {
+          "gateway": "https://ipfs.eth.aragon.network/ipfs"
+        }
+      },
       "registry": "0x98Df287B6C145399Aaa709692c8D308357bC085D",
+      "wsRPC": "wss://rinkeby.eth.aragon.network/ws",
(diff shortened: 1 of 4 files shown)


0xGabi:
@0x6431346e Do you think in this case we should also remove the `apm ipfs rpc` option? I'm hesitating because templates don't have UI, so should propagate quick.


Daniel comment, for reference: https://github.com/aragon/aragon-react-boilerplate/pull/61#discussion_r294216436

0x6431346e:
> @0x6431346e Do you think in this case we should also remove the `apm ipfs rpc` option? I'm hesitating because templates don't have UI, so should propagate quick.
> 
> Daniel comment, for reference: [aragon/aragon-react-boilerplate#61 (comment)](https://github.com/aragon/aragon-react-boilerplate/pull/61#discussion_r294216436)

Yes, let's out for this instead: https://github.com/aragon/aragon-cli/issues/535

FWIW, it would never propagate the UI anyway because [`apm.js` requires only `manifest.json` and `arapp.json`](https://github.com/aragon/apm.js/blob/master/src/index.js#L83)

PR #86, opened 2019-07-03
Update Multisig and Democracy gas tests scripts to new arapp file

### LINKED ISSUE #80: Fix test gas scripts
Both multisig and democracy templates have a test gas script that are broken, they are importing a relative file `../index_local.js` that doesn't exist.

bingen:
See commit 72d7f1b.

Closes #80 

Actually I don't remember what those scripts were used for, maybe we could just delete them.


sohkai:
> Actually I don't remember what those scripts were used for, maybe we could just delete them.

We were testing the gas costs for various things like org creation. It may be worth running them again and updating the READMEs with the current kits :).

PR #87, opened 2019-07-04
Move AGP-1 kit to concierge repo

facuspagnuolo:
Continue cleaning up this repo in order to hold only useful kits we do want to provide to users, I propose moving the agp-1 kit to concierge since it is more of an isolated DAO structure instead of a reusable one. I already created a [PR](https://github.com/aragon/concierge/pull/10) to add agp-1 to concierge.


facuspagnuolo:
@sohkai exactly, that was one of the drivers :)

Issue #91, opened 2019-07-12
Improve democracy template tests


facuspagnuolo:
Current democracy template tests can only be run with geth. We should provide support for ganache as well. Additionally, there are multiple steps along the suite that can be simplified using some of the helpers provided in `test-helpers` and `templates-shared`.

sohkai:
In some way, they were designed to be used specifically on geth so we could run these E2E smoke tests against a real node, but I agree running them on ganache would be useful for development and then we can run longer, nightly checks on CI with a real node.

Issue #90, opened 2019-07-12
Improve multisig tests


facuspagnuolo:
Current multisig template tests can only be run with geth. We should provide support for ganache as well. Additionally, there are multiple steps along the suite that can be simplified using some of the helpers provided in `test-helpers` and `templates-shared`.

PR #93, opened 2019-07-12
Refactor templates

facuspagnuolo:
Fixes https://github.com/aragon/dao-kits/issues/78

This PR proposes a new structure for the `dao-kits` repo aiming to hold only those templates that we are going to support "in production". Therefore, some breaking changes are being introduced:
- Drop `base` sub-repo
- Drop `beta-base` sub-repo
- Drop `helpers` repo including `test-token-deployer`
- Drop `kit` terminology in favor of `template`
- Unify deployment scripts. In `master` we have two flavors for those, one deploys related dependencies (ens, apm, etc), and the other one doesn't. The new deployment script is much more customizable and detects whether it has to deploy or not related dependencies making more clear which steps a developer should follow in order to deploy a template.
- Drop all the publish-related commands. These seemed to be useful for a small particular case. If we do need publish scripts, we can build new and more robust ones.
- Drop `KitBase` and `BetaKitBase` contracts in favor of a shared `BaseTemplate` contract that provides all the required functionality to create a dao, create tokens, install apps, manage permissions, etc

The following list denotes all the pending tasks that we have to complete in order to consider this refactor complete, we'll probably continue working on them in separate PRs tho:
- Improve multisig template tests (#90)
- Improve democracy template tests (#91)
- Add bare template tests (#92)
- Add tests for `shared-templates` 
- Test refactored templates in Rinkeby
- Refactor gas measurement scripts


sohkai:
> Drop helpers repo including test-token-deployer

I don't mind this, but we should find somewhere else to put this sub package :).

shared/contracts/BaseTemplate.sol
```diff
@@ -0,0 +1,191 @@
...
+    MiniMeTokenFactory public miniMeFactory;
+    IFIFSResolvingRegistrar public aragonID;
+
+    event DeployDao(address dao);
+    event DeployToken(address token);
+    event InstalledApp(address appProxy, bytes32 appId);
+
+    constructor(DAOFactory _daoFactory, ENS _ens, MiniMeTokenFactory _miniMeFactory, IFIFSResolvingRegistrar _aragonID) public {
+        require(isContract(address(_ens)), "TEMPLATE_ENS_NOT_CONTRACT");
+        require(isContract(address(_daoFactory)), "TEMPLATE_DAO_FACTORY_NOT_CONTRACT");
```
sohkai: Perhaps for `aragonID` and `miniMeFactory` we can check if it's 0x0, otherwise require them to also be contracts.

Also, given that we'll be using this base in who-knows-how-big templates, we may want to optimize its bytecode size. We could put everything into one require with one error message here. It would also be nice to lift the error strings into constants :).

shared/contracts/BaseTemplate.sol
```diff
@@ -0,0 +1,191 @@
...
+    function _createDAO() internal returns (Kernel dao, ACL acl) {
+        dao = daoFactory.newDAO(this);
+        emit DeployDao(address(dao));
+        acl = ACL(dao.acl());
+        _createPermission(acl, dao, dao.APP_MANAGER_ROLE());
+    }
+
+    /* ACL */
+
+    function _createPermission(ACL acl, address app, bytes32 permission) internal {
```
sohkai: Perhaps we can add a `ToFactory` or `FromFactory` to these function names, to make it clear who the assignee is.

shared/contracts/BaseTemplate.sol
```diff
@@ -0,0 +1,191 @@
...
+    function _installVotingApp(Kernel dao, MiniMeToken token, uint64 support, uint64 acceptance, uint64 duration) internal returns (Voting) {
+        Voting voting = Voting(_installNonDefaultApp(dao, VOTING_APP_ID));
+        voting.initialize(token, support, acceptance, duration);
+        return voting;
+    }
+
+    function _createVotingPermissions(ACL acl, Voting voting, address grantee, address manager) internal {
+        acl.createPermission(grantee, voting, voting.MODIFY_QUORUM_ROLE(), manager);
+        acl.createPermission(grantee, voting, voting.MODIFY_SUPPORT_ROLE(), manager);
+    }
```
sohkai: We may also want to add helpers to install the Survey app (not sure if we have plans to use it in a production kit)

shared/contracts/BaseTemplate.sol
```diff
@@ -0,0 +1,191 @@
...
+
+    function _installNonDefaultApp(Kernel dao, bytes32 appId) internal returns (address) {
+        return _installApp(dao, appId, new bytes(0), false);
+    }
+
+    function _installDefaultApp(Kernel dao, bytes32 appId) internal returns (address) {
+        return _installApp(dao, appId, new bytes(0), true);
+    }
+
+    function _installApp(Kernel dao, bytes32 appId, bytes data, bool setDefault) internal returns (address) {
```
sohkai: Perhaps we can rename this `_installAppFromAPM()`.

Also, what do you think about directly using this in most places, at least in the base template (since we don't need great UX ourselves)? We'd optimize the base contract's size, and could save a bit of gas by directly doing the `initialize()` via the `data` too.

shared/contracts/BaseTemplate.sol
```diff
@@ -0,0 +1,191 @@
...
+        require(isContract(address(miniMeFactory)), "TEMPLATE_MINIME_FACTORY_NOT_PROVIDED");
+        MiniMeToken token = miniMeFactory.createCloneToken(MiniMeToken(address(0)), 0, name, 18, symbol, true);
+        emit DeployToken(address(token));
+        return token;
+    }
+
+    /* IDS */
+
+    function _registerID(string name, address owner) internal {
+        require(isContract(address(aragonID)), "TEMPLATE_ARAGON_ID_NOT_PROVIDED");
```
sohkai: Since both the `registerID()` and `createToken()` will likely get called a lot, we can optimize their gas by changing the require to `require(address(aragonID) !== address(0), ERROR_CONSTANT)` and only checking if it's a contract in the constructor.

shared/contracts/BaseTemplate.sol
```diff
@@ -0,0 +1,191 @@
...
+        aragonID.register(keccak256(abi.encodePacked(name)), owner);
+    }
+}
```
sohkai: Nitpick, what do you think about keeping the prefixed `_` convention for argument names in this contract and the others, so it's the same as in other repos?

shared/helpers/apps.js
```diff
@@ -0,0 +1,19 @@
+const { hash: namehash } = require('eth-ens-namehash')
+
+const APPS = [
+  { id: 'agent', contractName: 'Agent' },
```
sohkai: It may be better calling the `id` `name` instead, or potentially changing it to be the `appName` (e.g. `agent.aragonpm.eth`) to be more consistent with the rest of the stack.

shared/lib/ArappFile.js
```diff
@@ -0,0 +1,44 @@
+const fs = require('fs')
```
sohkai: Nitpick on the file name, we generally use lowercase names unless we're exporting a class

shared/lib/ArappFile.js
```diff
@@ -0,0 +1,44 @@
...
+      const network = await getNetworkName()
+      const data = await ArappFile.file()
+      data.path = `contracts/${contractName}.sol`
+      if (data.environments === undefined) data.environments = {}
+      data.environments[network] = { appName, address, network, registry }
+      fs.writeFileSync(await ArappFile.filePath(), JSON.stringify(data, null, 2))
+    }
+  }
+
+  return ArappFile
```
sohkai: In these modules, I find the extra level of scoping a bit redundant, since we never need to access `this`.

Another problem that _could_ happen is someone accidentally changing a declaration at runtime (e.g. requiring `ArappFile` and then modifying `ArappFile.file` by accident).

What do you think about declaring them separately and only exporting them at the end?

```js
module.exports = web3 => {
  const { isLocalNetwork, getNetworkName } = require('./Network')(web3)

  async fileName() {
    return (await isLocalNetwork()) ? LOCAL_FILE_NAME : FILE_NAME
  }

  async filePath() {
    return path.resolve(await fileName())
  }

  ...

  return {
    fileName,
    filePath,
  }
}
```

shared/lib/TemplatesDeployer.js
```diff
@@ -0,0 +1,204 @@
+const { APPS } = require('../helpers/apps')
+const { hash: namehash } = require('eth-ens-namehash')
+
+const logDeploy = require('@aragon/os/scripts/helpers/deploy-logger')
+const deployAPM = require('@aragon/os/scripts/deploy-apm')
+const deployENS = require('@aragon/os/scripts/deploy-test-ens')
+const deployAragonID = require('@aragon/id/scripts/deploy-beta-aragonid')
+const deployDAOFactory = require('@aragon/os/scripts/deploy-daofactory')
```
sohkai: We've been shifting away from bundling these deployment scripts as part of individual repos, but I'm interested in what you think about our current strategy:

- Remove deployment scripts from most Aragon-related repos, unless they're optional dependencies in a general Aragon environment (e.g. the test tokens)
- Place the deployment scripts in [aragen](https://github.com/aragon/aragen), which should be able to spin up either a new chain, or deploy a base Aragon environment to an existing chain

We may want to start changing the script dependencies to be those in aragen, or we can wait until we finalize the migration (longer term).

shared/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "private": true,
+  "description": "Shared helpers for Aragon templates",
+  "dependencies": {
+    "@aragon/apps-agent": "1.0.0",
+    "@aragon/apps-finance": "3.0.0",
+    "@aragon/apps-shared-migrations": "1.0.0",
+    "@aragon/apps-shared-minime": "1.0.0",
+    "@aragon/apps-token-manager": "2.1.0",
+    "@aragon/apps-vault": "4.1.0",
+    "@aragon/apps-voting": "2.1.0",
```
sohkai: What do you think about not pinning these app dependencies, but using a `^`?

Since this is shared, we should use a looser range and let the actual templates pick which version of the apps they want to use. `@aragon/id` and `@aragon/os` should be outside the scope of other templates so we can pin them.

Originally we were going to set up something like dependabot via these loose versions to make sure the dao-templates still worked with new versions of any dependent apps. I think that still makes sense, and in production deployment commits we can either pin or include a `package-lock.json`.

shared/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "@aragon/test-helpers": "^2.0.0",
+    "@aragon/id": "2.0.3",
+    "@aragon/os": "4.2.0",
+    "dotenv": "^6.1.0",
```
sohkai: This is also in the devDependencies

shared/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "eth-ens-namehash": "^2.0.8",
+    "web3-eth-abi": "1.0.0-beta.33",
```
sohkai: Let's bump this to beta.37 (the rest of the repos will be following shortly)

templates/bare/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "@aragon/apps-vault": "^4.1.0",
+    "@aragon/apps-voting": "^2.1.0",
+    "@aragon/apps-finance": "^3.0.0",
+    "@aragon/apps-token-manager": "^2.1.0",
+    "@aragon/apps-shared-minime": "^1.0.0",
+    "@aragon/templates-shared": "^1.0.0"
+  },
+  "devDependencies": {
+    "ganache-cli": "^6.0.3",
+    "solidity-coverage": "0.5.8",
```
sohkai: Might as well bump this to v0.6 here and elsewhere :)

templates/bare/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "solium": "^1.0.4",
```
sohkai: We should migrate to `ethlint` (same as solium, but I think solium has stopped receiving updates)

templates/democracy/contracts/DemocracyTemplate.sol
```diff
@@ -0,0 +1,90 @@
...
+        _removePermission(acl, tokenManager, tokenManager.MINT_ROLE());
+
+        // Set up permissions
+        _createVaultPermissions(acl, vault, finance, voting);
+        _createFinancePermissions(acl, finance, voting, voting);
+        _createTokenManagerPermissions(acl, tokenManager, voting, voting);
+        _createEvmScriptsRegistryPermissions(acl, voting, voting);
+        _createVotingPermissions(acl, voting, tokenManager);
+        _transferPermission(acl, voting, dao, dao.APP_MANAGER_ROLE());
+        _transferPermission(acl, voting, acl, acl.CREATE_PERMISSIONS_ROLE());
```
sohkai: I imagine these last two will be done at the end of every template, what do you think about including something like `_transferRootPermissions()`?

I'm starting to think a good way to categorize these two "sensitive" permissions is to consider them "root" level permissions in an org.

templates/democracy/contracts/DemocracyTemplate.sol
```diff
@@ -0,0 +1,90 @@
...
+
+        _registerID(id, dao);
+    }
+
+    function _createVotingPermissions(ACL acl, Voting voting, TokenManager tokenManager) internal {
```
sohkai: This overloading is a bit scary, what do you think about naming it something else?

templates/multisig/contracts/MultisigTemplate.sol
```diff
@@ -0,0 +1,87 @@
...
+        require(requiredSignatures <= signers.length, "MULTISIG_INVALID_REQUIRED_SIGNATURES");
+
+        // We are subtracting 1 because comparison in Voting app is strict,
+        // while Multisig needs to allow equal too. So for instance in 2 out of 4
+        // multisig, we would define 50 * 10 ^ 16 - 1 instead of just 50 * 10 ^ 16,
+        // so 2 signatures => 2 * 10 ^ 18 / 4 = 50 * 10 ^ 16 > 50 * 10 ^ 16 - 1 would pass
+        // We can avoid safemath checks here as it's very unlikely a user will pass in enough
+        // signers to cause this to overflow
+        uint256 multiSigSupport = requiredSignatures * 10 ** 18 / signers.length - 1;
+        MiniMeToken token = _popTokenCache(msg.sender);
```
sohkai: Small optimization, let's move this to before `multiSigSupport`

templates/trust/contracts/TrustTemplate.sol
```diff
@@ -114,19 +95,30 @@ contract TrustKit is KitBase, IsContract {
...
+
+        _createVaultPermissions(acl, vault, finance, holdVoting);
+        _createFinancePermissions(acl, finance, holdVoting, holdVoting);
+        _createTokenManagerPermissions(acl, holdTokenManager, holdVoting);
+        _createTokenManagerPermissions(acl, heirsTokenManager, heirsVoting);
+        _createEvmScriptsRegistryPermissions(acl, holdVoting, holdVoting);
+        _createVotingPermissions(acl, holdTokenManager, holdVoting);
+        _createVotingPermissions(acl, heirsTokenManager, heirsVoting);
+        _createAgentPermission(acl, agent, agent.EXECUTE_ROLE(), holdVoting, heirsVoting);
+        _createAgentPermission(acl, agent, agent.RUN_SCRIPT_ROLE(), holdVoting, heirsVoting);
```
sohkai: Could we group these Voting and Agent permissions into their own custom functions in this template?

templates/democracy/contracts/DemocracyTemplate.sol
bingen: Hm, scary why? What could happen? I actually kind of like it, as the base one is missing `CREATE_VOTES_ROLE`. Plus the signature is different.

shared/helpers/assertRole.js
```diff
@@ -1,4 +1,5 @@
-module.exports = web3 => async (acl, app, manager, appName, roleName, grantee = manager) => {
+module.exports = web3 => async (acl, app, manager, roleName, grantee = manager) => {
```
bingen: Won't this have the problem that we can't differentiate two instances of the same app?

bingen: Besides, it would be nice to move Multisig and Democracy to use this instead of `checkRole`.

templates/democracy/arapp.json
```diff
@@ -2,34 +2,34 @@
...
       "registry": "0x5f6f7e8cc7346a11ca2def8f827b7a0b612c56a1",
-      "appName": "democracy-kit.aragonpm.eth",
+      "appName": "democracy-template.aragonpm.eth",
       "network": "rpc"
     },
     "rinkeby": {
       "registry": "0x98Df287B6C145399Aaa709692c8D308357bC085D",
       "wsRPC": "wss://rinkeby.eth.aragon.network/ws",
-      "appName": "democracy-kit.aragonpm.eth",
+      "appName": "democracy-template.aragonpm.eth",
```
bingen: We'll have to re-deploy these ones, right?

shared/helpers/assertRole.js
bingen: Ok, I see now this has been done in other PRs, forget it then :wink:

templates/democracy/contracts/DemocracyTemplate.sol
sohkai: > What could happen?

You think you're using the base one, but actually not. Or more likely, somewhere else down the line, someone thinks they're using the base one, but they get this version instead.

templates/democracy/arapp.json
sohkai: Yes, absolutely. Everything will need to be re-audited and re-deployed.

templates/democracy/contracts/DemocracyTemplate.sol
bingen: > You think you're using the base one, but actually not. 

But that's always a problem with overloading, right? Besides here the signature is different, so it makes it less likely.

> I find it primarily scary here because it's not intended to be overloaded; it's not really extensible functionality in my eyes.

As I said before, I disagree with this, because permissions setting in the first one is incomplete, so it makes total sense to extend it with the ones missing.

But of course perfectly fine for me if you prefer to rename it.

templates/democracy/contracts/DemocracyTemplate.sol
```diff
@@ -0,0 +1,90 @@
...
+    string private constant ERROR_MISSING_TOKEN_CACHE = "DEMOCRACY_MISSING_TOKEN_CACHE";
+    string private constant ERROR_INVALID_HOLDERS_STAKES_LEN = "DEMOCRACY_INVALID_HOLDERS_STAKES_LEN";
+
+    mapping (address => address) internal tokenCache;
+
+    constructor(DAOFactory _daoFactory, ENS _ens, MiniMeTokenFactory _miniMeFactory, IFIFSResolvingRegistrar _aragonID)
+        BaseTemplate(_daoFactory, _ens, _miniMeFactory, _aragonID)
+        public
+    {
+        // solium-disable-previous-line no-empty-blocks
```
sohkai: Here and in the Multisig kit, we may want to check `require(isContract(address(_fac.regFactory())));`

templates/democracy/readme.md
```diff
@@ -43,8 +41,8 @@ The network details will be automatically selected by the `arapp.json`'s environ
 
 ## Gas usage
 
-Tested running `GAS_REPORTER=true truffle test --network devnet test/gas.js`, plus `deploy-kit.js` script in `beta-base`.
+Tested running `GAS_REPORTER=true truffle test --network devnet test/gas.js`, plus `deploy.js` script.
 
-- Create the Kit:      2816197
+- Create template:     2816197
 - Create new token:    1738117
 - Deploy new instance: 5690035
```
sohkai: It would be nice to run these again and have updated values, to evaluate how the refactor will affect users.

facuspagnuolo:
Thanks @bingen and @sohkai for your feedback! I'll merge all these PRs into `next` and push one more PR at the end with all the corrections to avoid having lots of conflicts :)

lerna.json
```diff
@@ -1,8 +1,8 @@
 {
   "lerna": "2.4.0",
   "packages": [
-    "kits/*",
-    "helpers/*"
+    "templates/*",
+    "shared"
```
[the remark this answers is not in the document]
facuspagnuolo: actually `shared` is a single sub-repo

.travis.yml
```diff
@@ -1,25 +1,27 @@
...
-    - kits/democracy/node_modules
-    - kits/multisig/node_modules
+    - templates/shared/node_modules
+    - templates/democracy/node_modules
+    - templates/multisig/node_modules
+    - templates/trust/node_modules
 notifications:
   email: false
 node_js:
   - '9'
```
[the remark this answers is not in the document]
facuspagnuolo: Let's use 10

shared/contracts/BaseTemplate.sol
facuspagnuolo: I wasn't checking the `aragonID` and `miniMeFactory` cause they may not be used. Instead, we are doing a lazy check when trying to use them, i.e. before creating a token and before registering an ID.

shared/package.json
facuspagnuolo: I forgot to remove this one

templates/democracy/contracts/DemocracyTemplate.sol
facuspagnuolo: In other templates I used the `_createCustom...Permissions`, will apply that to all the templates

templates/trust/contracts/TrustTemplate.sol
facuspagnuolo: yup, I was waiting to merge some other PRs cause I was planning to leverage other helpers 😏

templates/democracy/readme.md
facuspagnuolo: will create a new issue for the gas measurements

templates/democracy/arapp.json
facuspagnuolo: yup

shared/contracts/BaseTemplate.sol
```diff
@@ -0,0 +1,191 @@
...
+        acl.removePermissionManager(app, permission);
+    }
+
+    function _transferPermission(ACL acl, address to, address app, bytes32 permission) internal {
+        acl.grantPermission(to, app, permission);
+        acl.revokePermission(address(this), app, permission);
+        acl.setPermissionManager(to, app, permission);
+    }
+
+    /* AGENT */
```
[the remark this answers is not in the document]
facuspagnuolo: Sure, I'll create a new issue for the inline doc

shared/contracts/BaseTemplate.sol
sohkai: > Instead, we are doing a lazy check when trying to use them

I'd rather do the expensive check (the `isContract()` check) on initialization than on each token creation and ID registration. If we check in the constructor that it's either `0x0` or a contract, then afterwards we can just make sure its not `0x0` when doing either of those actions.

PR #96, opened 2019-07-15
Bare: Add tests

### LINKED ISSUE #92: Add tests for the bare template

facuspagnuolo:

Fixes #92


templates/bare/test/bare.js
```diff
@@ -0,0 +1,52 @@
...
+const ACL = artifacts.require('ACL')
+const Kernel = artifacts.require('Kernel')
+const Voting = artifacts.require('Voting')
+const BareTemplate = artifacts.require('BareTemplate')
+
+contract('Bare', ([_, deployer, tokenAddress, authorized]) => {
+  let bareTemplate, dao, acl, voting, CREATE_VOTES_ROLE
+
+  const SUPPORT = 50e16
+  const ACCEPTANCE = 20e16
```
sohkai: Perhaps we could use `pct16()` (although perhaps we could also think of a better name for this utility, like `votePercentage()`)

PR #97, opened 2019-07-16
Company: Add template

facuspagnuolo:

Provides new template described [here](https://github.com/aragonone/json-templates/blob/master/company.json)


templates/company/contracts/CompanyTemplate.sol
```diff
@@ -0,0 +1,94 @@
...
+
+    constructor(DAOFactory _daoFactory, ENS _ens, MiniMeTokenFactory _miniMeFactory, IFIFSResolvingRegistrar _aragonID)
+        BaseTemplate(_daoFactory, _ens, _miniMeFactory, _aragonID)
+        public
+    {
+        // solium-disable-previous-line no-empty-blocks
+    }
+
+    function newTokenAndInstance(string id, address[] holders, uint256[] stakes) public {
+        require(holders.length == stakes.length, ERROR_INVALID_HOLDERS_STAKES_LEN);
```
bingen: This is already done below in `newInstance`.

templates/company/contracts/CompanyTemplate.sol
```diff
@@ -0,0 +1,94 @@
...
+        acl.createPermission(voting, voting, voting.MODIFY_QUORUM_ROLE(), voting);
+        acl.createPermission(voting, voting, voting.MODIFY_SUPPORT_ROLE(), voting);
+    }
+
+    function _createTokenManagerPermissions(ACL acl, TokenManager tokenManager, Voting voting) internal {
+        acl.createPermission(voting, tokenManager, tokenManager.BURN_ROLE(), voting);
+        acl.createPermission(voting, tokenManager, tokenManager.MINT_ROLE(), voting);
+    }
+
+    function _cacheToken(MiniMeToken token, address owner) internal {
```
bingen: I wonder if these 2 functions could be moved to `BaseTemplate` or to a library for them to be reused. Seems like a common use case.

templates/company/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Permissions
+
+| App               | Permission            | Grantee       | Manager |
+|-------------------|-----------------------|---------------|---------|
+| Kernel            | APP_MANAGER           | Voting        | Voting  |
+| ACL               | CREATE_PERMISSIONS    | Voting        | Voting  |
+| Voting            | CREATE_VOTES          | Token Manager | Voting  |
+| Voting            | MODIFY_QUORUM         | Voting        | Voting  |
+| Voting            | MODIFY_SUPPORT        | Voting        | Voting  |
```
sohkai: Just to note that this is a change from the current Democracy template, where we burn this role to disallow people from changing the support %.

lkngtn: Yes, this was definitely intentional. 

I don't see a good reason why support % should have ever been discarded.

templates/company/contracts/CompanyTemplate.sol
facuspagnuolo: All the caches work differently in the different templates, I don't think it could be reusable

facuspagnuolo:
@bingen I think @lkngtn already answered your questions in the corresponding threads. Regarding the optimization, I do agree on doing that but doing it now would be an immature refactor. Let's wait to iterate these templates a bit more and delay those changes

templates/company/contracts/CompanyTemplate.sol
bingen: Well, I would say these 4 are quite similar :wink: : 
https://github.com/aragon/dao-templates/blob/next/templates/company/contracts/CompanyTemplate.sol#L83
https://github.com/aragon/dao-templates/blob/next/templates/democracy/contracts/DemocracyTemplate.sol#L79
https://github.com/aragon/dao-templates/blob/next/templates/membership/contracts/MembershipTemplate.sol#L87
https://github.com/aragon/dao-templates/blob/next/templates/multisig/contracts/MultisigTemplate.sol#L76
https://github.com/aragon/dao-templates/blob/next/templates/reputation/contracts/ReputationTemplate.sol

And I guess there will eventually be more like these ones.

PR #98, opened 2019-07-16
Company Board: Add template

facuspagnuolo:

Provides new template described [here](https://github.com/aragonone/json-templates/blob/master/company-with-board.json)


shared/contracts/BaseTemplate.sol
```diff
@@ -65,20 +65,39 @@ contract BaseTemplate is APMNamehash, IsContract {
         acl.createPermission(address(this), app, permission, address(this));
     }
 
+    function _createPermissions(ACL acl, address[] grantees, address app, bytes32 permission, address manager) internal {
+        acl.createPermission(grantees[0], app, permission, address(this));
+        for (uint256 i = 1; i < grantees.length; i++) {
+            acl.grantPermission(grantees[i], app, permission);
+        }
+        acl.revokePermission(address(this), app, permission);
```
bingen: It seems slightly confusing to me that this function works different than the previous one here (template's permission removed).

templates/company-board/README.md
```diff
@@ -0,0 +1,51 @@
...
+| Share Voting        | MODIFY_QUORUM         | Share Voting        | Share Voting  |
+| Share Voting        | MODIFY_SUPPORT        | Share Voting        | Share Voting  |
+| Agent               | TRANSFER              | Finance             | Share Voting  |
+| Agent               | RUN_SCRIPT            | Board Voting        | Share Voting  |
+| Agent               | RUN_SCRIPT            | Share Voting        | Share Voting  |
+| Agent               | EXECUTE               | Board Voting        | Share Voting  |
+| Agent               | EXECUTE               | Share Voting        | Share Voting  |
+| Finance             | CREATE_PAYMENTS       | Board Voting        | Share Voting  |
+| Finance             | CREATE_PAYMENTS       | Share Voting        | Share Voting  |
+| Finance             | EXECUTE_PAYMENTS      | Share Voting        | Share Voting  |
```
bingen: I don't fully get this. These 3 permissions are granted to both voting apps? What's the idea, that any of those groups can approve those actions?

templates/company-board/README.md
```diff
@@ -0,0 +1,51 @@
+# Aragon Company Board template
+
+## Usage
+
+```
+company.prepareInstance(id)
+company.setupInstance(boardHolders, shareHolders, shareStakes)
```
bingen: Shouldn't this be `newTokensAndInstance` instead? Where is `newTokens` called otherwise?

templates/company-board/README.md
```diff
@@ -0,0 +1,51 @@
...
+## Permissions
+
+| App                 | Permission            | Grantee             | Manager       |
+|---------------------|-----------------------|---------------------|---------------|
+| Kernel              | APP_MANAGER           | Board Voting        | Share Voting  |
+| ACL                 | CREATE_PERMISSIONS    | Board Voting        | Share Voting  |
+| Board Voting        | CREATE_VOTES          | Board Token Manager | Share Voting  |
+| Board Voting        | MODIFY_QUORUM         | Share Voting        | Share Voting  |
+| Board Voting        | MODIFY_SUPPORT        | Share Voting        | Share Voting  |
+| Share Voting        | CREATE_VOTES          | Board Token Manager | Share Voting  |
```
bingen: So Share holders have all the power (like being Permission manager of everything, or being able to mint and burn Board tokens) but through their voting app, which votes can only be created by board members. So if board members collude they could keep all the effective power, right?

templates/company-board/contracts/CompanyBoardTemplate.sol
```diff
@@ -0,0 +1,168 @@
...
+        require(boardMembers.length > 0, ERROR_MISSING_BOARD_MEMBERS);
+        require(shareHolders.length == shareStakes.length, ERROR_INVALID_HOLDERS_STAKES_LEN);
+
+        newTokens();
+        prepareInstance(id);
+        setupInstance(boardMembers, shareHolders, shareStakes);
+    }
+
+    function newTokens() public {
+        MiniMeToken boardToken = _createToken(BOARD_TOKEN_NAME, BOARD_TOKEN_SYMBOL);
```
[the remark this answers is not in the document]
lkngtn: Definitely these should match, its not clear to me what the standard should be for a membership token (no decimals, 1 whole token per account) or (18 decimals, with one fractional unit max per account). The former seems more intuitive to me, but could understand if sticking with 18 decimals makes sense...

templates/company-board/README.md
lkngtn: Yup, both shareholders and the board have direct permissions to make finance transfers.

templates/company-board/README.md
lkngtn: If 100 percent of board members collude then yes they can prevent shareholders from doing anything, but if just one board member is honest then it should not be an issue. We could make it more permissive and allow any shareholder to create a vote... or use the tollgate app to allow for an "emergency" shareholder vote... but I think in most cases the assumption that there is atleast one honest board member is reasonable and by restricting vote creation to them it is easier for the organization to implement off-chain processes for how votes should or should not be created, how shareholders should be notified, etc. (the reason being that a board member who does not follow the process to create a vote can be removed). 

Anyways that was the rationale, hopefully it makes sense. :)

templates/company-board/contracts/CompanyBoardTemplate.sol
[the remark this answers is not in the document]
facuspagnuolo: I already check that with @lkngtn, we finally agreed on using 18 decimals, we should update the json file

shared/contracts/BaseTemplate.sol
facuspagnuolo: not using the remove permission helper here since we don't want to do that actually, we just want to remove the permission for the template, but to keep the one granted to the grantees, thus we are revoking the template's permission and then changing the manager

templates/company-board/contracts/CompanyBoardTemplate.sol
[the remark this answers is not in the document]
facuspagnuolo: Changing it to simply 1 token per address

shared/contracts/BaseTemplate.sol
```diff
@@ -65,15 +65,28 @@ contract BaseTemplate is APMNamehash, IsContract {
         acl.createPermission(address(this), app, permission, address(this));
     }
 
+    function _createPermissions(ACL acl, address[] grantees, address app, bytes32 permission, address manager) internal {
```
sohkai: Perhaps it would be nice to make this function more explicit, like `_createPermissionsAndSetManager`

shared/contracts/BaseTemplate.sol
```diff
@@ -65,15 +65,28 @@ contract BaseTemplate is APMNamehash, IsContract {
...
     function _removePermission(ACL acl, address app, bytes32 permission) internal {
         acl.revokePermission(address(this), app, permission);
         acl.removePermissionManager(app, permission);
     }
 
     function _transferPermission(ACL acl, address to, address app, bytes32 permission) internal {
+        _transferPermission(acl, to, to, app, permission);
+    }
+
+    function _transferPermission(ACL acl, address to, address manager, address app, bytes32 permission) internal {
```
sohkai: Perhaps it would be nice to make this more explicit, like `_transferPermissionWithManager`

templates/company-board/README.md
```diff
@@ -0,0 +1,51 @@
...
+| Agent               | EXECUTE               | Share Voting        | Share Voting  |
+| Finance             | CREATE_PAYMENTS       | Board Voting        | Share Voting  |
+| Finance             | CREATE_PAYMENTS       | Share Voting        | Share Voting  |
+| Finance             | EXECUTE_PAYMENTS      | Share Voting        | Share Voting  |
+| Finance             | DISABLE_PAYMENTS      | Share Voting        | Share Voting  |
+| Board Token Manager | MINT                  | Share Voting        | Share Voting  |
+| Board Token Manager | BURN                  | Share Voting        | Share Voting  |
+| Share Token Manager | MINT                  | Share Voting        | Share Voting  |
+| Share Token Manager | BURN                  | Share Voting        | Share Voting  |
+| EVMScriptRegistry   | REGISTRY_MANAGER      | Share Voting        | Share Voting  |
```
sohkai: Here and elsewhere, it would be good to group the `EVMScriptRegistry` with the `Kernel` and `ACL` since they're all "system apps"

templates/company-board/README.md
sohkai: Shareholders are still the role managers and have some access to functionality, so they could technically grant themselves various powers (minting a new board member, granting themselves access to create votes, etc.) if they needed to.

templates/company-board/README.md
sohkai: This definitely seems a bit sketchy; does the board need access to finance transfers and agent executes (I'd be OK with safe executes where no token transfers are allowed)?

templates/company-board/test/company-board.js
```diff
@@ -0,0 +1,185 @@
...
+      finance = Finance.at(financeEvents[0].args.proxy)
+      boardVoting = Voting.at(votingEvents[0].args.proxy)
+      shareVoting = Voting.at(votingEvents[1].args.proxy)
+      boardTokenManager = TokenManager.at(tokenManagerEvents[0].args.proxy)
+      shareTokenManager = TokenManager.at(tokenManagerEvents[1].args.proxy)
+    })
+
+    it('costs less than 6e6 each tx', async () => {
+      assert.isAtMost(tokenReceipt.receipt.gasUsed, 3.5e6, 'tokens creation should cost almost 3.5e6 gas')
+      assert.isAtMost(prepareReceipt.receipt.gasUsed, 1.7e6, 'prepare DAO should cost almost 1.7e6 gas')
```
sohkai: I'm wondering if we can group the token creation and initial org creation into one step, so this template just takes two transactions to finish?

templates/company-board/README.md
bingen: > ~Shareholders are still the role managers and have some access to functionality, so they could technically grant themselves various powers (minting a new board member, granting themselves access to create votes, etc.) if they needed to.~
> 
> Nevermind, I see shareholder votes can only be created by board token holders.

Yes, that was actually my concern.

Although I guess that the assumption that at least one board member is honest, I don't fully get the problem of assigning `CREATE_VOTE_ROLE` to shareholders too.

shared/contracts/BaseTemplate.sol
bingen: Yes, but I'm not talking about the helper (sorry my explanation was awful, haha). I mean that after calling `_createPermission` the template will hold the permission, while after calling `_createPermissions` (with a very similar name) it won't. That's what I find confusing.

bingen: I see that names were changed in #102 , now it's less confusing.

templates/company-board/README.md
facuspagnuolo: `newTokens` will be merged into `prepareInstance`

templates/company-board/README.md
sohkai: Mostly spam, probably. Typically shareholders do not get to create votes themselves, unless they are a fairly large holder (at which point the board is most likely theirs anyway).

You'd typically have another set of people, the management, that would also have control over creating votes / budgets, and that could be introduced via another token (if not an entire sub organization with its own finance app and etc.).

PR #100, opened 2019-07-17
Reputation: Add template

facuspagnuolo:
**CHANGE BASE BRANCH TO MASTER AFTER #93 GETS MERGED**

Provides new template described [here](https://github.com/aragonone/json-templates/blob/master/reputation.json)

PR #95, opened 2019-07-15
Democracy: Improve tests

facuspagnuolo:
**CHANGE BASE BRANCH TO MASTER AFTER #93 GETS MERGED**

Fixes #91

PR #102, opened 2019-07-22
Polish new templates and templates refactor

facuspagnuolo:
Addressing feedback from @bingen and @sohkai related to #93, #94, #95, #96, #97, #98, #99, #100 

Some pending tweaks I'll create issues to address them are:
- Providing support for the newest apps: survey and payroll
- Provide gas measure scripts and measure current costs
- Write inline documentation
- Improve geth `ethereum-data` handling
- Optimize apps installation using initialize data
- Provide helper to mint tokens


templates/bare/package.json
```diff
@@ -26,10 +26,12 @@
   },
   "devDependencies": {
     "@aragon/test-helpers": "^2.0.0",
+    "eth-ens-namehash": "^2.0.8",
+    "eth-gas-reporter": "^0.1.1",
     "ganache-cli": "^6.0.3",
-    "solidity-coverage": "0.5.8",
     "solium": "^1.0.4",
     "truffle": "4.1.14",
+    "truffle-extract": "^1.2.1",
```
sohkai: Not sure we use this dependency (we don't extract the ABIs, and I'm not sure if we'll ever publish these templates to npm)

templates/company-board/README.md
```diff
@@ -2,12 +2,20 @@
...
+
 ```
-company.prepareInstance(id)
-company.setupInstance(boardHolders, shareHolders, shareStakes)
+template.prepareInstance(name)
+template.setupInstance(boardHolders, shareHolders, shareStakes)
 ```
 
-- `id`: Identifier to be registered in AragonID
+- `name`: Name for org, will assign `[name].aragonid.eth` (check capitalization)
```
sohkai: I'm fairly confident (not 100%) that ENS names, like domain names, are not case-sensitive.

templates/democracy/contracts/DemocracyTemplate.sol
```diff
@@ -4,87 +4,90 @@ import "@aragon/templates-shared/contracts/BaseTemplate.sol";
 
 
 contract DemocracyTemplate is BaseTemplate {
+    bool constant private TOKEN_TRANSFERABLE = false;
```
sohkai: I think this should be true?

templates/multisig/contracts/MultisigTemplate.sol
```diff
@@ -7,81 +7,89 @@ import "@aragon/templates-shared/contracts/BaseTemplate.sol";
...
+        MiniMeToken token = _createToken(_name, _symbol);
         _cacheToken(token, msg.sender);
         return token;
     }
 
-    function newInstance(string id, address[] signers, uint256 requiredSignatures) public {
-        require(signers.length > 0, "MULTISIG_EMPTY_SIGNERS");
-        require(requiredSignatures > 0, "MULTISIG_REQUIRED_SIGNATURES_ZERO");
-        require(requiredSignatures <= signers.length, "MULTISIG_INVALID_REQUIRED_SIGNATURES");
+    function newInstance(string _id, address[] _signer, uint256 _requiredSignatures) public {
```
sohkai: Nitpick: should we keep the `_signers` variable name here and in `newTokenAndInstance`, since it's an array?

templates/trust/contracts/TrustTemplate.sol
```diff
@@ -77,182 +83,182 @@ contract TrustTemplate is BaseTemplate {
         return dao;
     }
 
-    function setupDAO(string id, address[] beneficiaryKeys, address[] heirs, uint256[] heirsStake) public returns (Kernel) {
-        require(_hasDaoCache(msg.sender), ERROR_MISSING_CACHE_TOKENS_FOR_SENDER);
-        require(heirs.length == heirsStake.length, ERROR_BAD_HEIRS_LENGTH);
-        require(beneficiaryKeys.length == BENEFICIARY_KEYS_AMOUNT, ERROR_BAD_BENEFICIARY_KEYS_LENGTH);
-        uint256 blockedHeirsSupply = _calculateBlockedHeirsSupply(heirsStake);
+    function setupDAO(string _id, address[] _beneficiaryKeys, address[] _heirs, uint256[] _heirsStake) public returns (Kernel) {
```
sohkai: Nitpick: what do you think about `heirsStakes` instead, to keep the "noun" plural?

templates/trust/contracts/TrustTemplate.sol
```diff
@@ -77,182 +83,182 @@ contract TrustTemplate is BaseTemplate {
...
         _createFinancePermissions(acl, finance, holdVoting, holdVoting);
-        _createTokenManagerPermissions(acl, holdTokenManager, holdVoting);
-        _createTokenManagerPermissions(acl, heirsTokenManager, heirsVoting);
         _createEvmScriptsRegistryPermissions(acl, holdVoting, holdVoting);
-        _createVotingPermissions(acl, holdTokenManager, holdVoting);
-        _createVotingPermissions(acl, heirsTokenManager, heirsVoting);
-        _createAgentPermission(acl, agent, agent.EXECUTE_ROLE(), holdVoting, heirsVoting);
-        _createAgentPermission(acl, agent, agent.RUN_SCRIPT_ROLE(), holdVoting, heirsVoting);
+        _createCustomVotingPermissions(acl, holdTokenManager, holdVoting);
+        _createCustomAgentPermissions(acl, agent, holdVoting, heirsVoting);
```
sohkai: It's a bit odd the Agent permissions are done in-between the Voting permission assignments

shared/contracts/BaseTemplate.sol
```diff
@@ -56,155 +62,178 @@ contract BaseTemplate is APMNamehash, IsContract {
         dao = daoFactory.newDAO(this);
         emit DeployDao(address(dao));
         acl = ACL(dao.acl());
-        _createPermission(acl, dao, dao.APP_MANAGER_ROLE());
+        _createPermissionForTemplate(acl, dao, dao.APP_MANAGER_ROLE());
```
sohkai: It may be worth adding a comment saying that the template now has both the `kernel.APP_MANAGER_ROLE` and `acl.CREATE_PERMISSIONS_ROLE`, and that at the end they should use `_transferRootPermissionsFromTemplate()`

shared/contracts/BaseTemplate.sol
```diff
@@ -56,155 +62,178 @@ contract BaseTemplate is APMNamehash, IsContract {
...
-    function _removePermission(ACL acl, address app, bytes32 permission) internal {
-        acl.revokePermission(address(this), app, permission);
-        acl.removePermissionManager(app, permission);
+    function _transferPermissionFromTemplate(ACL _acl, address _to, address _app, bytes32 _permission) internal {
+        _transferPermissionFromTemplate(_acl, _to, _to, _app, _permission);
     }
 
-    function _transferPermission(ACL acl, address to, address app, bytes32 permission) internal {
-        _transferPermission(acl, to, to, app, permission);
+    function _transferPermissionFromTemplate(ACL _acl, address _to, address _manager, address _app, bytes32 _permission) internal {
```
sohkai: What do you think about an API like

```
function _transferPermissionAndManagerFromTemplate(ACL _acl, address _to, address _app, bytes32 _permission, address _manager) internal {
```

instead?

shared/contracts/BaseTemplate.sol
```diff
@@ -56,155 +62,178 @@ contract BaseTemplate is APMNamehash, IsContract {
...
     }
 
-    function _transferPermission(ACL acl, address to, address manager, address app, bytes32 permission) internal {
-        acl.grantPermission(to, app, permission);
-        acl.revokePermission(address(this), app, permission);
-        acl.setPermissionManager(manager, app, permission);
+    function _transferRootPermissionsFromTemplate(Kernel _dao, address _to) internal {
+        ACL _acl = ACL(_dao.acl());
+        _transferPermissionFromTemplate(_acl, _to, _dao, _dao.APP_MANAGER_ROLE());
+        _transferPermissionFromTemplate(_acl, _to, _acl, _acl.CREATE_PERMISSIONS_ROLE());
```
sohkai: I'm not sure we would ever use this version when transferring root permissions; setting the Kernel and ACL as the manager of the roles is basically setting them to `BURN_ROLE`.

shared/contracts/BaseTemplate.sol
```diff
@@ -56,155 +62,178 @@ contract BaseTemplate is APMNamehash, IsContract {
...
-    function _registerID(string name, address owner) internal {
-        require(isContract(address(aragonID)), "TEMPLATE_ARAGON_ID_NOT_PROVIDED");
-        aragonID.register(keccak256(abi.encodePacked(name)), owner);
+    function _registerID(string _name, address _owner) internal {
+        require(address(aragonID) != address(0), ERROR_ARAGON_ID_NOT_PROVIDED);
+        aragonID.register(keccak256(abi.encodePacked(_name)), _owner);
+    }
+
+    function _ensureAragonIdIsValid(address _aragonID) internal view {
+        require(isContract(address(_aragonID)), ERROR_ARAGON_ID_NOT_CONTRACT);
```
sohkai: Here and in `_ensureMiniMeFactoryIsValid()`, we should allow `address(0)` as well (`isContract` will return false if `address(0)` is given).

shared/contracts/BaseTemplate.sol
```diff
@@ -56,155 +62,178 @@ contract BaseTemplate is APMNamehash, IsContract {
...
         agent.initialize();
+        // We assume that installing the Agent app as default is in order to replace the Vault app which is
+        // normally installed as default. Thus, we are setting its ID as the Vault id that the Kernel will use.
+        _dao.setRecoveryVaultAppId(AGENT_APP_ID);
         return agent;
     }
 
-    function _installNonDefaultAgentApp(Kernel dao) internal returns (Agent) {
-        Agent agent = Agent(_installNonDefaultApp(dao, AGENT_APP_ID));
+    function _installNonDefaultAgentApp(Kernel _dao) internal returns (Agent) {
```
sohkai: Rather than two functions, does it make sense just to add the "default"-ness as an argument flag instead?

shared/contracts/BaseTemplate.sol
```diff
@@ -56,155 +62,178 @@ contract BaseTemplate is APMNamehash, IsContract {
...
         acl = ACL(dao.acl());
-        _createPermission(acl, dao, dao.APP_MANAGER_ROLE());
+        _createPermissionForTemplate(acl, dao, dao.APP_MANAGER_ROLE());
     }
 
     /* ACL */
 
-    function _createPermission(ACL acl, address app, bytes32 permission) internal {
-        acl.createPermission(address(this), app, permission, address(this));
+    function _createPermissions(ACL _acl, address[] _grantees, address _app, bytes32 _permission, address _manager) internal {
```
sohkai: One thing I'm not sure about is on the argument order for these base functions.

It's true that the ACL has its arguments in order of `entity`, `app`, and `role`, but that's because it's in the context of the ACL (and not a template).

I wonder if it would make more sense (or maybe it's more confusing?) if we re-ordered the arguments in the templates as `acl`, `app`, `role`, `grantee/s`, `manager`, so that `acl`, `app`, and `role` are consistently in the same order at the start.

templates/company/test/company.js
```diff
@@ -1,8 +1,9 @@
 const { hash: namehash } = require('eth-ens-namehash')
```
sohkai: Whereever we add gas tests, it would be nice to also include one for `newTokenAndInstance()` (and the like that combine multiple steps into one).

templates/bare/package.json
facuspagnuolo: We don't but it was in all the templates with their corresponding npm scripts, I didn't want to remove it in case it was useful for someone

templates/company-board/README.md
facuspagnuolo: Sure. I brought that comment from the democracy readme and I thought it may be useful advice.

templates/democracy/contracts/DemocracyTemplate.sol
facuspagnuolo: Good catch! Will add some tests for that as well 👍

templates/multisig/contracts/MultisigTemplate.sol
facuspagnuolo: Sure, I might have messed it up when renaming the variables

shared/contracts/BaseTemplate.sol
facuspagnuolo: Not sure what you mean but it is actually being used by the `BareTemplate` to gran root permissions to a root address passed by the user.

shared/contracts/BaseTemplate.sol
facuspagnuolo: These helpers are used actually to check that those dependencies are valid: non-zero and a contract. All the templates using them have to check that, otherwise, their creation methods will revert.

sohkai: Ah got it, that makes sense.

shared/contracts/BaseTemplate.sol
sohkai: But I think it uses the other one, that includes the `manager` parameter, right? This one sets the managers to the ACL and Kernel.

templates/company-board/README.md
sohkai: From https://github.com/ethereum/EIPs/blob/master/EIPS/eip-137.md#name-syntax:

> Note that while upper and lower case letters are allowed in names, the UTS46 normalisation process case-folds labels before hashing them, so two names with different case but identical spelling will produce the same namehash.

So I think we're safe to remove this comment :)

templates/company/test/company.js
facuspagnuolo: sure, will do

templates/company-board/README.md
```diff
@@ -2,12 +2,20 @@
...
+template.newTokens()
+```
+
+Create a new company with board entity:
+
 ```
-company.prepareInstance(id)
-company.setupInstance(boardHolders, shareHolders, shareStakes)
+template.prepareInstance(name)
+template.setupInstance(boardHolders, shareHolders, shareStakes)
```
bingen: I think this hasn't been answered:

https://github.com/aragon/dao-templates/pull/98/files#r305383821

facuspagnuolo: Answered in the thread :)

shared/contracts/BaseTemplate.sol
facuspagnuolo: Agreed with @bingen, I like the idea, but I prefer consistency. WDYT about replicating that in the `ACL` @sohkai ?

shared/contracts/BaseTemplate.sol
facuspagnuolo: Sorry, I meant that this one is being used by all the templates except the `BareTemplate`... This helper sets the `_to` address as the manager, note that it is calling the following function for both roles:

```sol
_transferPermissionFromTemplate(ACL _acl, address _to, address _app, bytes32 _permission)
```

However, I'll reuse the other helper used by the `BareTemplate` to make it clearer

shared/contracts/BaseTemplate.sol
sohkai: We should avoid changes to the ACL's API for the foreseeable future, until there's a _really_ good reason to change it, so let's keep the order as is (`entity`, `app`, `role`, `manager`).

templates/multisig/test/multisig.js
```diff
@@ -26,283 +27,363 @@ const ExecutionTarget = artifacts.require('ExecutionTarget')
...
         })
 
-        it('fails trying to modify support threshold directly', async () => {
-          await assertRevert(voting.changeSupportRequiredPct.request(REQUIRED_SUPPORT.add(1), { from: owner }), 'APP_AUTH_FAILED')
+        it('costs ~6.6e6 gas', async () => {
+          if (creationStyle === 'single') {
+            assert.isAtMost(instanceReceipt.receipt.gasUsed, 6.7e6, 'create script should cost almost 6.7e6 gas')
+          } else if (creationStyle === 'separate') {
+            assert.isAtMost(tokenReceipt.receipt.gasUsed, 1.8e6, 'create token script should cost almost 1.8e6 gas')
+            assert.isAtMost(instanceReceipt.receipt.gasUsed, 5e6, 'create instance script should cost almost 5e6 gas')
```
sohkai: Interesting, this is a good sign, at least. The old kit used to require 5.7M gas to do just the instance (no token), so it looks like we've shaved off a bit with the new base template 👍

shared/contracts/BaseTemplate.sol
```diff
@@ -43,168 +49,194 @@ contract BaseTemplate is APMNamehash, IsContract {
...
+        Finance finance = Finance(_installNonDefaultApp(_dao, FINANCE_APP_ID));
+        finance.initialize(_vault, _periodDuration);
         return finance;
     }
 
-    function _createFinancePermissions(ACL acl, Finance finance, address grantee, address manager) internal {
-        acl.createPermission(grantee, finance, finance.CREATE_PAYMENTS_ROLE(), manager);
-        acl.createPermission(grantee, finance, finance.EXECUTE_PAYMENTS_ROLE(), manager);
-        acl.createPermission(grantee, finance, finance.MANAGE_PAYMENTS_ROLE(), manager);
+    function _createFinancePermissions(ACL _acl, Finance finance, address _grantee, address _manager) internal {
```
sohkai:
```suggestion
    function _createFinancePermissions(ACL _acl, Finance _finance, address _grantee, address _manager) internal {
```

shared/contracts/BaseTemplate.sol
facuspagnuolo: I gave it a try and it increased the bytecode length. The `TrustTemplate` is hitting the limit (see [here](https://travis-ci.org/aragon/dao-templates/jobs/563044874)), honestly, I prefer the syntax-sugar rather than dealing with that now 😅

facuspagnuolo: Actually, the previous commit is already exceeding the limit, don't get why it passed on the CI 🤷‍♀

bingen: Hm, interesting. I did a quick check and I see big differences:

| Contract | Size |
|-------------|-------|
| Trust | 24625 |
| Democracy | 11835 |
| Multisig | 12019 |
| Company | 12188 |
| C. Board | 15350 |
| Reputation | 12045 |
| Membership | 11953 |

Any idea why? what does this Trust one have that makes it twice as big as most of the other ones?

facuspagnuolo: I think once we optimize apps installation with a single call instead of two we will have more space for this change

facuspagnuolo: @bingen yea, probably because of the Gnosis multisig

shared/contracts/BaseTemplate.sol
```diff
@@ -39,10 +39,10 @@ contract BaseTemplate is APMNamehash, IsContract {
...
     string constant private ERROR_MINIME_FACTORY_NOT_CONTRACT = "TEMPLATE_MINIME_FAC_NOT_CONTRACT";
 
-    ENS public ens;
-    DAOFactory public daoFactory;
-    MiniMeTokenFactory public miniMeFactory;
-    IFIFSResolvingRegistrar public aragonID;
+    ENS internal ens;
+    DAOFactory internal daoFactory;
+    MiniMeTokenFactory internal miniMeFactory;
+    IFIFSResolvingRegistrar internal aragonID;
```
sohkai: Hmm, I do like having these stay `public`; I do use them quite a bit when debugging / figuring out what a template is attached to (although I guess I could look up the contract creation transaction, but that's a bit more involved).

PR #116, opened 2019-07-24
Bare: Fix template name for ropsten

facuspagnuolo:


sohkai:
Hmm, not sure we should do this. If we redeploy, we should either find a new name for template, or publish it as v2.

facuspagnuolo:
I already removed all the other deployed addresses and was planning to populate these files after the deploy. These files are already stale at the moment. WDYT?

sohkai:
Ahh, I see where the confusion lies.

Those addresses are **not** stale (and that configuration could already be added to each template's `arapp.json`); they're simply the ENS registry's address that we point the CLI to when we publish. Because the "appName" is just an ENS name, the CLI can fetch the repo if it knows which ENS registry to use, making all of this automatic.

facuspagnuolo:
@sohkai thx for clarifying, let's fix the template name for ropsten then :)

PR #119, opened 2019-07-25
Allows to pass more parameters to templates

### LINKED ISSUE #111: Parameterize new templates app settings
New templates (reputation, membership, and both company templates) do not allow the user to specify some variables that should be customizable. These are:
- voting duration
- voting support
- voting acceptance
- finance period duration (use 0 for 30 days default)
- token name
- token symbol

ajsantander:
:checkered_flag:  *Ready for Review* :checkered_flag:
Fix #111 

#### Tasks (In reputation, membership, company and company-board templates):
- [x] Token name and symbol
- [x] Vote duration
- [x] Vote support
- [x] Vote acceptance
- [x] Finance period

### DIFF
--- a/templates/company-board/contracts/CompanyBoardTemplate.sol
+++ b/templates/company-board/contracts/CompanyBoardTemplate.sol
@@ -58,7 +58,9 @@ contract CompanyBoardTemplate is BaseTemplate {
         uint64 _shareVoteDuration,
         uint64 _boardSupportRequired,
         uint64 _shareSupportRequired
-    ) public {
+    ) 
+        public 
+    {
         (Agent agent, Finance finance) = _setupCommon();
         (TokenManager boardTokenManager, Voting boardVoting) = _setupBoard(_boardMembers, _boardVoteDuration, _boardSupportRequired);
         (TokenManager shareTokenManager, Voting shareVoting) = _setupShare(_shareHolders, _shareStakes, _shareVoteDuration, _shareSupportRequired);
@@ -109,7 +111,9 @@ contract CompanyBoardTemplate is BaseTemplate {
         TokenManager _boardTokenManager,
         Voting _shareVoting,
         TokenManager _shareTokenManager
-    ) internal {
+    ) 
+        internal 
+    {
         (Kernel dao,,) = _getCache(msg.sender);
         ACL acl = ACL(dao.acl());
         _createVaultPermissions(acl, Vault(_agent), _finance, _shareVoting);
--- a/templates/company-board/contracts/CompanyBoardTemplate.sol
+++ b/templates/company-board/contracts/CompanyBoardTemplate.sol
@@ -13,6 +13,8 @@ contract CompanyBoardTemplate is BaseTemplate {
     uint64 constant private FINANCE_PERIOD = uint64(30 days);
 
     bool constant private BOARD_TRANSFERABLE = false;
+    string constant private BOARD_TOKEN_NAME = "Board Token";	
+    string constant private BOARD_TOKEN_SYMBOL = "BOARD";
     uint8 constant private BOARD_TOKEN_DECIMALS = uint8(0);
     uint256 constant private BOARD_MAX_PER_ACCOUNT = uint256(1);
 
@@ -44,9 +46,9 @@ contract CompanyBoardTemplate is BaseTemplate {
         _ensureMiniMeFactoryIsValid(_miniMeFactory);
     }
 
-    function prepareInstance(string _boardTokenName, string _boardTokenSymbol, string _shareTokenName, string _shareTokenSymbol) public {
+    function prepareInstance(string _shareTokenName, string _shareTokenSymbol) public {
         (Kernel dao,) = _createDAO();
-        MiniMeToken boardToken = _createToken(_boardTokenName, _boardTokenSymbol, BOARD_TOKEN_DECIMALS);
+        MiniMeToken boardToken = _createToken(BOARD_TOKEN_NAME, BOARD_TOKEN_SYMBOL, BOARD_TOKEN_DECIMALS);
         MiniMeToken shareToken = _createToken(_shareTokenName, _shareTokenSymbol, SHARE_TOKEN_DECIMALS);
         _storeCache(dao, boardToken, shareToken, msg.sender);
     }
--- a/templates/company-board/contracts/CompanyBoardTemplate.sol
+++ b/templates/company-board/contracts/CompanyBoardTemplate.sol
@@ -9,7 +9,6 @@ contract CompanyBoardTemplate is BaseTemplate {
     string constant private ERROR_MISSING_SHARE_MEMBERS = "COMPANY_MISSING_SHARE_MEMBERS";
     string constant private ERROR_BAD_HOLDERS_STAKES_LEN = "COMPANY_BAD_HOLDERS_STAKES_LEN";
 
-    uint64 constant private ONE_PCT = uint64(1e16);
     uint64 constant private FINANCE_PERIOD = uint64(30 days);
 
     bool constant private BOARD_TRANSFERABLE = false;
@@ -22,14 +21,12 @@ contract CompanyBoardTemplate is BaseTemplate {
     uint8 constant private SHARE_TOKEN_DECIMALS = uint8(18);
     uint256 constant private SHARE_MAX_PER_ACCOUNT = uint256(0);
 
-    uint64 constant private BOARD_MIN_ACCEPTANCE_QUORUM = uint64(40 * ONE_PCT);   // 40%
-
-    uint64 constant private SHARE_MIN_ACCEPTANCE_QUORUM = uint64(5 * ONE_PCT);    // 5%
-
     struct Cache {
         address dao;
         address boardToken;
         address shareToken;
+        address agent;
+        address finance;
     }
 
     mapping (address => Cache) internal cache;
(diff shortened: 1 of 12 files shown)


templates/company-board/contracts/CompanyBoardTemplate.sol
```diff
@@ -48,10 +44,10 @@ contract CompanyBoardTemplate is BaseTemplate {
         _ensureMiniMeFactoryIsValid(_miniMeFactory);
     }
 
-    function prepareInstance() public {
+    function prepareInstance(string _boardTokenName, string _boardTokenSymbol, string _shareTokenName, string _shareTokenSymbol) public {
```
facuspagnuolo: We shouldn't parameterize the board token, see https://github.com/aragon/dao-templates/issues/111#issuecomment-514400907

ajsantander: Totally missed that comment. Thx for the reminder!
*Updated* so that CompanyBoardTemplate only has the shareToken's name and symbol as parameters.

templates/company-board/contracts/CompanyBoardTemplate.sol
```diff
@@ -48,42 +39,90 @@ contract CompanyBoardTemplate is BaseTemplate {
...
+        string _id, 
+        address[] _boardMembers, 
+        address[] _shareHolders, 
+        uint256[] _shareStakes,
+        uint64 _boardVoteDuration,
+        uint64 _shareVoteDuration,
+        uint64 _boardSupportRequired,
+        uint64 _shareSupportRequired,
+        uint64 _boardMinAcceptanceQuorum,
+        uint64 _shareMinAcceptanceQuorum
```
facuspagnuolo: I realized that we could use 2 lists of 3 `uint64` elements each to specify the voting app's parameters. This would allow us to reduce the number of declared variables and maybe we can move the finance and agent apps setup to this step, which will reduce gas costs since we won't need to cache them

templates/company/contracts/CompanyTemplate.sol
```diff
@@ -30,48 +24,66 @@ contract CompanyTemplate is BaseTemplate {
...
+        address[] _holders, 
+        uint256[] _stakes, 
+        string _tokenName, 
+        string _tokenSymbol,
+        uint64 _voteDuration,
+        uint64 _supportRequired,
+        uint64 _minAcceptanceQuorum,
+        uint64 _financePeriod
+    ) 
+        public
```
facuspagnuolo: Not tied to these changes, but can we make all the `newTokenAndInstance` functions external?

templates/company-board/contracts/CompanyBoardTemplate.sol
ajsantander: Nize!
Applied. It does make the code much cleaner, but it doesn't save gas :(

templates/company/contracts/CompanyTemplate.sol
ajsantander: Done. Note that no changes were made to templates that do not use the newToken/newTokenAndInstance function pair.

templates/company-board/contracts/CompanyBoardTemplate.sol
```diff
@@ -48,42 +38,92 @@ contract CompanyBoardTemplate is BaseTemplate {
...
+    ) 
+        public 
+    {
+        require(_boardVoteSettings.length == 3, ERROR_BAD_VOTE_SETTINGS);
+        require(_shareVoteSettings.length == 3, ERROR_BAD_VOTE_SETTINGS);
+        (Agent agent, Finance finance) = _setupCommon(_financePeriod);
+        (TokenManager boardTokenManager, Voting boardVoting) = _setupBoard(_boardMembers, _boardVoteSettings[0], _boardVoteSettings[1], _boardVoteSettings[2]);
+        (TokenManager shareTokenManager, Voting shareVoting) = _setupShare(_shareHolders, _shareStakes, _shareVoteSettings[0], _shareVoteSettings[1], _shareVoteSettings[2]);
+        _setupPermissions(boardVoting, boardTokenManager, shareVoting, shareTokenManager, agent, finance);
+        _registerDAO(_id);
```
facuspagnuolo: I'm sure we can optimize the caching here a little bit. Probably you can pop up the cache at the beginning and pass the dao/acl by parameter to the rest of the functions instead of loading the cache on each sub function

templates/company/contracts/CompanyTemplate.sol
facuspagnuolo: Probably not yet for this transaction, but the total should have been reduced since we are not storing the finance and agent apps any more

templates/company-board/contracts/CompanyBoardTemplate.sol
facuspagnuolo: Probably not yet for this transaction, but the total should have been reduced since we are not storing the finance and agent apps any more

templates/company-board/contracts/CompanyBoardTemplate.sol
ajsantander: Yeah, I had tried this before but it "caused stack too deep" errors. I uploaded that version in case you want to see. I failed at the linked line: https://github.com/aragon/dao-templates/blob/parameterize_templates_simpler_cache/templates/company-board/contracts/CompanyBoardTemplate.sol#L65

ajsantander:
> Looking good, left another comment on how we can optimize the cache for the company-board template... WDYT about replicating the voting settings thing for the rest of the templates?

TBH, I think that using an array for all voting settings is less explicit in function interfaces that dedicated parameters per setting. I wouldn't use it unless it's necessary. Now, if we put consistency before all other things, yes, we'd have to do it, but it would make the other templates just a bit less clear.

templates/company-board/contracts/CompanyBoardTemplate.sol
facuspagnuolo: Ofc that indexing variables at that level won't help, you should pass the array to the internal function and check it there

ajsantander: Yes, you're right! Still, I just tried changing that, and the stack too deep error persists. I've pushed it to that temporary branch nonetheless, so that we may use this code as reference for a future gas optimization.

templates/company-board/contracts/CompanyBoardTemplate.sol
```diff
@@ -8,29 +8,19 @@ contract CompanyBoardTemplate is BaseTemplate {
...
     string constant private ERROR_MISSING_SHARE_MEMBERS = "COMPANY_MISSING_SHARE_MEMBERS";
     string constant private ERROR_BAD_HOLDERS_STAKES_LEN = "COMPANY_BAD_HOLDERS_STAKES_LEN";
-
-    uint64 constant private ONE_PCT = uint64(1e16);
-    uint64 constant private FINANCE_PERIOD = uint64(30 days);
+    string constant private ERROR_BAD_VOTE_SETTINGS = "COMPANY_BAD_VOTE_SETTINGS";
 
     bool constant private BOARD_TRANSFERABLE = false;
-    string constant private BOARD_TOKEN_NAME = "Board Token";
+    string constant private BOARD_TOKEN_NAME = "Board Token";
```
sohkai: Note that there's some extra whitespace lingering in most of the contract files.

(LMK if your editor doesn't have an easy way to remove them, I use [this vim command](https://github.com/bronson/vim-trailing-whitespace) to fix it automatically)

templates/company-board/test/company-board.js
```diff
@@ -43,59 +53,111 @@ contract('Company with board', ([_, owner, boardMember1, boardMember2, shareHold
...
 
     it('costs ~10.4e6 gas', async () => {
-      assert.isAtMost(prepareReceipt.receipt.gasUsed, 5e6, 'prepare script should cost almost 5e6 gas')
-      assert.isAtMost(setupReceipt.receipt.gasUsed, 5.4e6, 'setup script should cost almost 5.4e6 gas')
+      const prepareGas = prepareReceipt.receipt.gasUsed;
+      const setupGas = setupReceipt.receipt.gasUsed;
+      const totalGas = prepareGas + setupGas;
+      assert.isAtMost(prepareGas, 5.0e6, 'prepare script should cost almost 5.0e6 gas')
+      assert.isAtMost(setupGas, 5.4e6, 'setup script should cost almost 5.4e6 gas')
+      assert.isAtMost(totalGas, 10.4e6, 'prepare + setup scripts should cost almost 10.4e6 gas');
```
sohkai: Note on the trailing `;` too, generally we don't use them anymore :)

templates/company/contracts/CompanyTemplate.sol
```diff
@@ -30,48 +24,66 @@ contract CompanyTemplate is BaseTemplate {
...
 
-    function newTokenAndInstance(string _id, address[] _holders, uint256[] _stakes) public {
-        newToken();
-        newInstance(_id, _holders, _stakes);
+    function newTokenAndInstance(
+        string _id, 
+        address[] _holders, 
+        uint256[] _stakes, 
+        string _tokenName, 
+        string _tokenSymbol,
```
sohkai: Super small thing, but the parameter order _may_ be a little bit more obvious if we had the arguments in the same order they would be used in, i.e. `_tokenName, `_tokenSymbol`, `_id`, `_holders`, ...

That way a frontend could _conceivably_ do something like `creationArgs = [].concat(tokenArgs, instanceArgs)` and it'd just work (not that we would likely be this loose in our own frontend).

templates/company-board/contracts/CompanyBoardTemplate.sol
ajsantander: Oops! Thanks for catching that. I believe that @facuspagnuolo already fixed this in https://github.com/aragon/dao-templates/pull/123

I was planning on installing a linter to avoid this sort of stuff, but didn't find the time last week. Anyways, I installed that plugin for the mean time. Love it :D

templates/company-board/test/company-board.js
ajsantander: :+1: 
I'll also be setting up my linter for JS this week to comply with the team's JS coding style.

PR #120, opened 2019-07-25
Chore: Add publish scripts

facuspagnuolo:


facuspagnuolo:
> although perhaps we can also log something immediately before like "Deployed 'template-name' at:"

The problem is that we need to tail the address only, we will need to parse the output if we do that, or output it in multiple lines

PR #121, opened 2019-07-26
Toggles between Agent or Vault app

### LINKED ISSUE #112: Allow toggling between a Vault or Agent app
New templates introduce the usage of the Agent app in favor of the Vault app. We should allow the user to stay with the Vault app. We can introduce a new boolean to the templates interface to decide whether to install a Vault or an Agent app

ajsantander:
Fix #112 

#### Tasks
- [x] Reputation template
- [x] Membership template
- [x] Company template
- [x] Company board template

### DIFF
--- a/templates/company-board/test/company-board.js
+++ b/templates/company-board/test/company-board.js
@@ -13,6 +13,7 @@ const ENS = artifacts.require('ENS')
 const ACL = artifacts.require('ACL')
 const Kernel = artifacts.require('Kernel')
 const Agent = artifacts.require('Agent')
+const Vault = artifacts.require('Vault')
 const Voting = artifacts.require('Voting')
 const Finance = artifacts.require('Finance')
 const TokenManager = artifacts.require('TokenManager')
@@ -24,7 +25,7 @@ const ZERO_ADDRESS = '0x0000000000000000000000000000000000000000'
 
 contract('Company with board', ([_, owner, boardMember1, boardMember2, shareHolder1, shareHolder2, shareHolder3]) => {
   let daoID, template, dao, acl, ens, setupReceipt, prepareReceipt
-  let shareVoting, boardVoting, boardTokenManager, shareTokenManager, boardToken, shareToken, finance, agent
+  let shareVoting, boardVoting, boardTokenManager, shareTokenManager, boardToken, shareToken, finance, agent, vault
 
   const BOARD_MEMBERS = [boardMember1, boardMember2]
   const SHARE_HOLDERS = [shareHolder1, shareHolder2, shareHolder3]
@@ -36,14 +37,14 @@ contract('Company with board', ([_, owner, boardMember1, boardMember2, shareHold
     template = CompanyTemplate.at(address)
   })
 
-  before('build dao ID', () => {
-    daoID = randomId()
-  })
-
   context('when the creation fails', () => {
+    before('build dao ID', () => {
+      daoID = randomId()
+    })
+
     context('when there was no instance prepared before', () => {
       it('reverts', async () => {
-        await assertRevert(template.setupInstance.request(daoID, BOARD_MEMBERS, SHARE_HOLDERS, SHARE_STAKES), 'COMPANY_MISSING_DAO_CACHE')
+        await assertRevert(template.setupInstance.request(daoID, BOARD_MEMBERS, SHARE_HOLDERS, SHARE_STAKES, true), 'COMPANY_MISSING_DAO_CACHE')
       })
     })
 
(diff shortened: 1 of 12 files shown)


ajsantander:
@facuspagnuolo I only applied the changes to just one template (ReputationTemplate) so far, to see if you like the approach. I'll continue with the other templates once you confirm and/or we review if this is the way to go.

Notice how the implementation in the contract is trivial, but adds some complexity to the tests.

templates/reputation/test/reputation.js
```diff
@@ -39,162 +40,181 @@ contract('Reputation', ([_, owner, holder1, holder2]) => {
...
+          it('sets up DAO and ACL permissions correctly', async () => {
+            await assertRole(acl, dao, voting, 'APP_MANAGER_ROLE')
+            await assertRole(acl, acl, voting, 'CREATE_PERMISSIONS_ROLE')
+          })
+
+          it('sets up EVM scripts registry permissions correctly', async () => {
+            const reg = await EVMScriptRegistry.at(await acl.getEVMScriptRegistry())
+            await assertRole(acl, reg, voting, 'REGISTRY_ADD_EXECUTOR_ROLE')
+            await assertRole(acl, reg, voting, 'REGISTRY_MANAGER_ROLE')
+          })
```
facuspagnuolo: Instead of doing a `for` loop we can move the whole test suite that you wanna ensure for both cases to a function and call it twice, it will look cleaner. I'm thinking sth like
```js
for (const creationStyle of ['single', 'separate']) {
   const itHandlesInstanceCreationsProperly = useAgentAsVault => {
      it('test 1', async () => { ... })
      it('test 2', async () => { ... })
   }
   
   context('when using an agent as vault', () => {
      itHandlesInstanceCreationsProperly(true)
      
      it('should have agent app correctly setup', async () => { ... })
   })

   context('when using a regular vault', () => {
      itHandlesInstanceCreationsProperly(false)

      it('should have vault app correctly setup', async () => {...})
   })
}
```

ajsantander: Cool, refactored and implemented this style on all tests.

ajsantander:
Note that the CI is not passing for a known and expected reason: a stack too deep error that should be fixed once we rebase with the changes made in PR #119

templates/company-board/test/company-board.js
```diff
@@ -111,170 +116,203 @@ contract('Company with board', ([_, owner, boardMember1, boardMember2, shareHold
...
-    it('registers a new DAO on ENS', async () => {
-      const ens = ENS.at((await deployedAddresses()).registry)
-      const aragonIdNameHash = namehash(`${daoID}.aragonid.eth`)
-      const resolvedAddress = await PublicResolver.at(await ens.resolver(aragonIdNameHash)).addr(aragonIdNameHash)
-      assert.equal(resolvedAddress, dao.address, 'aragonId ENS name does not match')
-    })
+      before('load apps', async () => {
+        const installedApps = getInstalledAppsById(setupReceipt)
+        if(useAgentAsVault) assert.equal(installedApps.agent.length, 1, 'should have installed 1 agent app')
+        else assert.equal(installedApps.vault.length, 1, 'should have installed 1 vault app')
```
sohkai: A quick note on style, we generally use the form of:

```
if (condition) {
  ...
} else {
  ...
}
```

Unless the clause is _really_ simple.

In any case, at some point we will have prettier handle most of this formatting :).

templates/company/README.md
```diff
@@ -14,16 +14,15 @@ template.newToken()
...
+template.newInstance(name, holders, stakes, voteSettings, financePeriod, useAgentAsVault)
 ```
 
 - `name`: Name for org, will assign `[name].aragonid.eth`
 - `holders`: Array of token holder addresses
 - `stakes`: Array of token stakes for holders (token has 18 decimals, multiply token amount `* 10^18`)
-- `voteDuration`: Time period in which a vote can be resolved
-- `supportRequired`: Percentage of total voting power or minimum quorum that a vote needs to be approved/executed
-- `minAcceptanceQuorum`: Percentage of the total voting power to consider when calculating support
+- `voteSettings`: Array of [voteDuration, supportRequired, minAcceptanceQuorum] for votes by the organization's members
```
sohkai: For all the `voteSettings`, what do you think about changing the order so that it's similar to the `Voting.initialize()` parameter order (`supportRequiredPct, `minAcceptanceQuorumPct`, `voteTime`)?

ajsantander: Yep. Totally makes sense. I believe that this is already implemented by @facuspagnuolo in https://github.com/aragon/dao-templates/pull/123

PR #122, opened 2019-07-26
Adds helpers to mint tokens

### LINKED ISSUE #109: Provide helpers to mint tokens
All the templates we're providing implement this logic, we should be able to provide that from the `BaseTemplate` contract

ajsantander:
Fix #109 

#### Tasks
- [x] Reputation template
- [x] Membership template
- [x] Company template
- [x] Company board template
- [x] Trust template
- [x] Bare template

### DIFF
--- a/shared/contracts/BaseTemplate.sol
+++ b/shared/contracts/BaseTemplate.sol
@@ -157,6 +157,28 @@ contract BaseTemplate is APMNamehash, IsContract {
         _acl.createPermission(_grantee, _tokenManager, _tokenManager.REVOKE_VESTINGS_ROLE(), _manager);
     }
 
+    function _mintTokens(ACL _acl, TokenManager _tokenManager, address[] _holders, uint256[] _stakes) internal {
+        _createPermissionForTemplate(_acl, _tokenManager, _tokenManager.MINT_ROLE());
+        for (uint256 i = 0; i < _holders.length; i++) {
+            _tokenManager.mint(_holders[i], _stakes[i]);
+        }
+        _removePermissionFromTemplate(_acl, _tokenManager, _tokenManager.MINT_ROLE());
+    }
+
+    function _mintTokens(ACL _acl, TokenManager _tokenManager, address[] _holders, uint256 _stake) internal {
+        _createPermissionForTemplate(_acl, _tokenManager, _tokenManager.MINT_ROLE());
+        for (uint256 i = 0; i < _holders.length; i++) {
+            _tokenManager.mint(_holders[i], _stake);
+        }
+        _removePermissionFromTemplate(_acl, _tokenManager, _tokenManager.MINT_ROLE());
+    }
+
+    function _mintTokens(ACL _acl, TokenManager _tokenManager, address _holder, uint256 _stake) internal {
+        _createPermissionForTemplate(_acl, _tokenManager, _tokenManager.MINT_ROLE());
+        _tokenManager.mint(_holder, _stake);
+        _removePermissionFromTemplate(_acl, _tokenManager, _tokenManager.MINT_ROLE());
+    }
+
     /* VAULT */
 
     function _installVaultApp(Kernel _dao) internal returns (Vault) {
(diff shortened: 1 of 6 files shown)


shared/contracts/BaseTemplate.sol
```diff
@@ -225,6 +225,28 @@ contract BaseTemplate is APMNamehash, IsContract {
...
+            _tokenManager.mint(_holders[i], _stake);
+        }
+        _removePermissionFromTemplate(_acl, _tokenManager, _tokenManager.MINT_ROLE());
+    }
+
+    function _mintTokens(ACL _acl, TokenManager _tokenManager, address _holder, uint256 _stake) internal {
+        _createPermissionForTemplate(_acl, _tokenManager, _tokenManager.MINT_ROLE());
+        _tokenManager.mint(_holder, _stake);
+        _removePermissionFromTemplate(_acl, _tokenManager, _tokenManager.MINT_ROLE());
+    }
```
facuspagnuolo: WDYT about moving these fns to the "token manager" section?

templates/company/contracts/CompanyTemplate.sol
```diff
@@ -60,6 +60,7 @@ contract CompanyTemplate is BaseTemplate {
         TokenManager tokenManager = _installTokenManagerApp(dao, token, TOKEN_TRANSFERABLE, TOKEN_MAX_PER_ACCOUNT);
         Voting voting = _installVotingApp(dao, token, _voteSettings[1], _voteSettings[2], _voteSettings[0]);
 
+        // Mint tokens
         _mintTokens(acl, tokenManager, _holders, _stakes);
 
         _setupPermissions(dao, acl, agentOrVault, voting, finance, tokenManager, _useAgentAsVault);
```
sohkai: In this template we should remove the overriding `_mintTokens()` definition

shared/contracts/BaseTemplate.sol
```diff
@@ -157,6 +157,28 @@ contract BaseTemplate is APMNamehash, IsContract {
...
+
+    function _mintTokens(ACL _acl, TokenManager _tokenManager, address[] _holders, uint256 _stake) internal {
+        _createPermissionForTemplate(_acl, _tokenManager, _tokenManager.MINT_ROLE());
+        for (uint256 i = 0; i < _holders.length; i++) {
+            _tokenManager.mint(_holders[i], _stake);
+        }
+        _removePermissionFromTemplate(_acl, _tokenManager, _tokenManager.MINT_ROLE());
+    }
+
+    function _mintTokens(ACL _acl, TokenManager _tokenManager, address _holder, uint256 _stake) internal {
```
sohkai: I don't know if we should bother with this version of the command. If we do, I would also rename it to just `_mintToken` since all it can ever mint is one token to one address.

templates/trust/contracts/TrustTemplate.sol
```diff
@@ -108,15 +108,20 @@ contract TrustTemplate is BaseTemplate {
...
         Agent agent = _installNonDefaultAgentApp(_dao);
         Finance finance = _installFinanceApp(_dao, vault, 30 days);
         (Voting holdVoting, Voting heirsVoting, TokenManager holdTokenManager, TokenManager heirsTokenManager) = _installTokenApps(_dao);
 
-        _mintHoldTokens(acl, holdTokenManager, _beneficiaryKeys);
-        _mintHeirsTokens(acl, heirsTokenManager, _heirs, _heirsStakes, _blockedHeirsSupply);
+        // Mint tokens
+        _mintTokens(acl, holdTokenManager, _beneficiaryKeys, 1e18);
+        _mintTokens(acl, heirsTokenManager, _heirs, _heirsStakes);
+        _mintTokens(acl, heirsTokenManager, address(0), _blockedHeirsSupply);
```
sohkai: Note that this is fairly gas inefficient, since we assigning and re-assigning permissions on the `heirsTokenManager` here.

Although uglier, we could stuff this burned minting into the previous call instead.

templates/company/contracts/CompanyTemplate.sol
ajsantander: Yep. I noticed this stray function after we merged the PR and removed it here: https://github.com/aragon/dao-templates/commit/c67e697af8e0b2ea3c9e3654132bf024443bd10b

shared/contracts/BaseTemplate.sol
ajsantander: @sohkai 
1) I think that this function is handy because it encapsulates the complexity of having to grant permissions and then revoke them during the mint, so I would keep it.
2) Regarding the rename, the function only mints to one beneficiary address, yes, but it mints any amount of tokens, not just one, so you are actually minting "tokens" to that address.

templates/trust/contracts/TrustTemplate.sol
ajsantander: I _think_ I see your point. However, `heirsTokenManager` is not available in the previous call, se we can't move the mintint to the zero address to there. It only be comes available in this function, a few lines back at L116.

sohkai: Ah yes, I meant the previous call to `_mintTokens()`, directly above. We could add the `address(0), _blockedHeirsSupply` into a new array copied from `_heirsStakes` (although admittedly this is getting complex).

shared/contracts/BaseTemplate.sol
sohkai: Good call on the name, let’s leave it as an overload 👍 .

templates/trust/contracts/TrustTemplate.sol
ajsantander: Yep. That would indeed be more efficient, but quite obscure, don't you think? I suggest that we drop the conversation here (since its a closed PR) and create a new issue to address this inefficiency.

sohkai: Yes, agreed. Especially on a template that already requires quite a lot of gas, and may not be used all the time, I think this is acceptable for now (less so for something like the company template 😄 ).

PR #123, opened 2019-07-27
Company-Board: Optimize gas costs

facuspagnuolo:


templates/company-board/contracts/CompanyBoardTemplate.sol
```diff
@@ -38,97 +38,109 @@ contract CompanyBoardTemplate is BaseTemplate {
...
-        _shareVoting = _installVotingApp(dao, shareToken, _shareSupportRequired, _shareMinAcceptanceQuorum, _shareVoteDuration);
+        boardVoting = _installVotingApp(_dao, boardToken, _boardVotingSettings[0], _boardVotingSettings[1], _boardVotingSettings[2]);
+        shareVoting = _installVotingApp(_dao, shareToken, _shareVotingSettings[0], _shareVotingSettings[1], _shareVotingSettings[2]);
+        boardTokenManager = _installTokenManagerApp(_dao, boardToken, BOARD_TRANSFERABLE, BOARD_MAX_PER_ACCOUNT);
+        shareTokenManager = _installTokenManagerApp(_dao, shareToken, SHARE_TRANSFERABLE, SHARE_MAX_PER_ACCOUNT);
 
-        // Mint tokens
-        _mintTokens(acl, _shareTokenManager, _shareHolders, _shareStakes);
+        _mintTokens(acl, boardTokenManager, _boardMembers, 1);
+        _mintTokens(acl, shareTokenManager, _shareHolders, _shareStakes);
```
sohkai: Not a huge thing, but I don't really see the point of `_setupApps()` and `_installTokenApps()` if they could just be rolled into `setupInstance()`

templates/company-board/contracts/CompanyBoardTemplate.sol
```diff
@@ -38,97 +38,109 @@ contract CompanyBoardTemplate is BaseTemplate {
...
     }
 
     function _setupPermissions(
-        Voting _boardVoting, 
+        Kernel _dao,
```
sohkai: Maybe it make sense to take an `ACL` here, since by this time we'll have already gotten it.

templates/company-board/contracts/CompanyBoardTemplate.sol
```diff
@@ -172,17 +184,22 @@ contract CompanyBoardTemplate is BaseTemplate {
...
-        delete cache[_owner];
+    function _popTokensCache(address _owner) internal returns (MiniMeToken boardToken, MiniMeToken shareToken) {
+        Cache storage c = cache[_owner];
+        require(c.boardToken != address(0) && c.shareToken != address(0), ERROR_MISSING_CACHE);
+
+        boardToken = MiniMeToken(c.boardToken);
+        shareToken = MiniMeToken(c.shareToken);
+        delete c.boardToken;
+        delete c.shareToken;
     }
```
sohkai: In the same vein of rolling things up, I liked your suggestion of just popping everything at the start of `setupInstance()`.

templates/company/contracts/CompanyTemplate.sol
```diff
@@ -47,18 +47,18 @@ contract CompanyTemplate is BaseTemplate {
...
+        require(_votingSettings.length == 3, ERROR_BAD_VOTE_SETTINGS);
         MiniMeToken token = _popTokenCache(msg.sender);
 
         // Create DAO and install apps
         (Kernel dao, ACL acl) = _createDAO();
         Vault agentOrVault = _useAgentAsVault ? _installDefaultAgentApp(dao) : _installVaultApp(dao);
         Finance finance = _installFinanceApp(dao, agentOrVault, _financePeriod == 0 ? DEFAULT_FINANCE_PERIOD : _financePeriod);
         TokenManager tokenManager = _installTokenManagerApp(dao, token, TOKEN_TRANSFERABLE, TOKEN_MAX_PER_ACCOUNT);
-        Voting voting = _installVotingApp(dao, token, _voteSettings[1], _voteSettings[2], _voteSettings[0]);
+        Voting voting = _installVotingApp(dao, token, _votingSettings[0], _votingSettings[1], _votingSettings[2]);
```
sohkai: As mentioned elsewhere, let's use the `Voting.initialize()` parameter order here and in the BaseTemplate to avoid confusion.

templates/company-board/contracts/CompanyBoardTemplate.sol
```diff
@@ -38,97 +38,109 @@ contract CompanyBoardTemplate is BaseTemplate {
...
-        address[] _boardMembers, 
-        address[] _shareHolders, 
+        string _id,
+        address[] _boardMembers,
+        address[] _shareHolders,
         uint256[] _shareStakes,
-        uint64[] _boardVoteSettings, /* [voteDuration, supportRequired, minAcceptanceQuorum] */
-        uint64[] _shareVoteSettings, /* idem */
+        uint64[3] _boardVotingSettings, /* [voteDuration, supportRequired, minAcceptanceQuorum] */
+        uint64[3] _shareVotingSettings, /* [voteDuration, supportRequired, minAcceptanceQuorum] */
```
sohkai: I'm curious what solidity does with this extra information about the array's length. Does it enforce that it's only of length up to 3?

templates/company-board/contracts/CompanyBoardTemplate.sol
facuspagnuolo: I already tried that, we are getting `stack too deep` issues

templates/company-board/contracts/CompanyBoardTemplate.sol
facuspagnuolo: We haven't and if we do we get `stack too deep` issues... note that the function calling this is `_setupApps` which already receives 8 arguments and install 6 apps, handling memory variables there is a PITA

templates/company-board/contracts/CompanyBoardTemplate.sol
facuspagnuolo: same argument here, `stack too deep` issues

templates/company/contracts/CompanyTemplate.sol
facuspagnuolo: Note that this change was in favor of that, this is `support, acceptance, duration`

templates/company-board/contracts/CompanyBoardTemplate.sol
facuspagnuolo: You will get an `Out of bounds array access` compilation error if you try sth like `array[4]` somewhere... additionally, not sure if truffle or web3 do some kind of validation for calls

PR #124, opened 2019-07-29
Add payroll app to all templates

ajsantander:
:construction:  WIP :construction: 

#### Tasks
- [x] Add Payroll app functionality in base template
- [x] Add Payroll app in company template (review before applying to all others)
- [x] Company board template
- [x] Membership template
- [x] Reputation template


shared/helpers/abi.js
```diff
@@ -0,0 +1,27 @@
...
+    })
+  },
+
+  encodeFunctionCall: (funcSignature, types, params) => {
+    params = abiUtil.stringifyParams(params)
+    return abi.encodeFunctionSignature(funcSignature) + abi.encodeParameters(types, params).replace('0x', '');
+  }
+}
+
+module.exports = abiUtil
```
facuspagnuolo: Let's inline `abiUtil` to be consistent with the rest of the modules here (see other module exports for reference).

shared/helpers/abi.js
```diff
@@ -0,0 +1,27 @@
...
+          return p
+        }
+        else {
+          return `${p}`
+        }
+      }
+    })
+  },
+
+  encodeFunctionCall: (funcSignature, types, params) => {
```
facuspagnuolo: We are following this convention

```suggestion
  encodeFunctionCall(signature, types, params) {
```

shared/scripts/test-ganache.sh
```diff
@@ -19,7 +19,7 @@ cleanup() {
 setup_testing_variables() {
   PORT=${PORT-8545}
   BALANCE=${BALANCE-100000}
-  GAS_LIMIT=${GAS_LIMIT-8000000}
+  GAS_LIMIT=${GAS_LIMIT-10000000}
```
facuspagnuolo: We shouldn't be working with non-real gas values. If we are hitting a gas limit, we should split the transactions

facuspagnuolo:
Left a few comments, would you mind updating this branch with `master`?

shared/contracts/BaseTemplate.sol
```diff
@@ -128,6 +132,25 @@ contract BaseTemplate is APMNamehash, IsContract {
...
+    function _installPayrollApp(Kernel _dao, Finance _finance, address _denominationToken, IFeed _priceFeed, uint64 _rateExpiryTime) internal returns (Payroll) {
+        Payroll payroll = Payroll(_installNonDefaultApp(_dao, PAYROLL_APP_ID));
+        payroll.initialize(_finance, _denominationToken, _priceFeed, _rateExpiryTime);
+        return payroll;
+    }
+
+    function _createPayrollPermissions(ACL _acl, Payroll _payroll, address _grantee, address _manager) internal {
+        _acl.createPermission(_grantee, _payroll, _payroll.ADD_BONUS_ROLE(), _manager);
+        _acl.createPermission(_grantee, _payroll, _payroll.ADD_EMPLOYEE_ROLE(), _manager);
+        _acl.createPermission(_grantee, _payroll, _payroll.ADD_REIMBURSEMENT_ROLE(), _manager);
```
sohkai: We may want a different manager for the `ADD_BONUS`, `ADD_REIMBURSEMENT`, and `ADD_EMPLOYEE` roles.

The other ones are very powerful “setting” management roles, whereas these are directly targeted at managing employees. Perhaps we can have the concept of the “Payroll settings manager” (should be Voting or etc.) and the “Payroll employee manager” (could be a high privilege account)?

shared/contracts/BaseTemplate.sol
```diff
@@ -128,6 +132,26 @@ contract BaseTemplate is APMNamehash, IsContract {
...
+    /* PAYROLL */
+
+    function _installPayrollApp(Kernel _dao, Finance _finance, address _denominationToken, IFeed _priceFeed, uint64 _rateExpiryTime) internal returns (Payroll) {
+        Payroll payroll = Payroll(_installNonDefaultApp(_dao, PAYROLL_APP_ID));
+        payroll.initialize(_finance, _denominationToken, _priceFeed, _rateExpiryTime);
+        return payroll;
+    }
+
+    function _createPayrollPermissions(ACL _acl, Payroll _payroll, address _employeeManager, address _settingsManager, address _permissionsManager) internal {
+        _acl.createPermission(_employeeManager, _payroll, _payroll.ADD_BONUS_ROLE(), _permissionsManager);
```
sohkai: Let's leave some comments for this separation of roles

shared/contracts/BaseTemplate.sol
```diff
@@ -257,4 +281,12 @@ contract BaseTemplate is APMNamehash, IsContract {
     function _ensureAragonIdIsValid(address _aragonID) internal view {
         require(isContract(address(_aragonID)), ERROR_ARAGON_ID_NOT_CONTRACT);
     }
+
+    /* HELPERS */
+
+    // TODO: Move to helper (in this repo for now)
+    function _toAddress(uint256 _value) internal pure returns(address) {
+        require(_value <= uint160(-1), ERROR_CANNOT_CAST_VALUE_TO_ADDRESS);
+        return address(_value);
```
sohkai: Does solidity overflow if the value is too big?

shared/helpers/abi.js
```diff
@@ -0,0 +1,26 @@
...
+        return p
+      }
+      else {
+        return `${p}`
+      }
+    }
+  })
+}
+
+function encodeFunctionCall(funcSignature, types, params) {
```
sohkai: Can we not just use `web3-eth-abi`'s [`encodeFunctionCall()`](https://github.com/ethereum/web3.js/blob/1.x/packages/web3-eth-abi/src/index.js#L200) directly?

templates/trust/package.json
```diff
@@ -43,6 +43,7 @@
     "@aragon/apps-agent": "^1.0.0",
     "@aragon/apps-vault": "^4.1.0",
     "@aragon/apps-voting": "^2.1.0",
+    "@aragon/apps-payroll": "^1.0.0-rc.1",
```
sohkai: Don't think we need this dependency here 😄

shared/helpers/assertRevert.js
```diff
@@ -27,8 +27,8 @@ module.exports = web3 => {
     }
   }
 
-  async function assertRevertGeth(request, reason) {
-    const tx = request.params[0]
+  async function assertRevertGeth(truffleInstance, requestOrTx, reason) {
```
sohkai: It seems quite odd to require the truffle-contract instance here (and it deviates from a lot of other `assertRevert` code we have across the org), why was this needed?

templates/company/test/company.js
```diff
@@ -105,7 +123,15 @@ contract('Company', ([_, owner, holder1, holder2]) => {
...
             } else if (creationStyle === 'separate') {
               tokenReceipt = await template.newToken(TOKEN_NAME, TOKEN_SYMBOL, { from: owner })
-              instanceReceipt = await template.newInstance(daoID, HOLDERS, STAKES, VOTING_SETTINGS, DEFAULT_FINANCE_PERIOD, useAgentAsVault, { from: owner })
+              if (installPayroll) {
+                const dummyPayrollFeed = template.address
+                const employeeManager = useOwnerAsEmployeeManager ? owner : ZERO_ADDRESS
+                const payrollSettings = [PAYROLL_DENOMINATION_TOKEN, dummyPayrollFeed, PAYROLL_RATE_EXPIRY_TIME, employeeManager]
+                instanceReceipt = await newInstance(daoID, HOLDERS, STAKES, VOTING_SETTINGS, DEFAULT_FINANCE_PERIOD, useAgentAsVault, payrollSettings)
+              }
+              else {
```
sohkai: Super style nitpick, we generally put `else`s on the trailing bracket

Prefer

```js
if {
} else {
}
```

Over

```js
if {
}
else {
}
```

templates/company/test/company.js
```diff
@@ -243,6 +277,63 @@ contract('Company', ([_, owner, holder1, holder2]) => {
...
+        if (creationStyle === 'separate') {
+          context('when installing the payroll', () => {
+
+            context('when using the voting app for employee management', () => {
+              itHandlesInstanceCreationsProperly(true, true, false)
+
+              it('should have payroll app correctly setup', async () => {
+                assert.isTrue(await payroll.hasInitialized(), 'payroll not initialized')
+                assert.equal(await payroll.denominationToken(), PAYROLL_DENOMINATION_TOKEN)
+                assert.equal(await payroll.feed(), template.address)
```
sohkai: Perhaps we could use a constant for the feed address too, in the relevant contexts.

templates/company/contracts/CompanyTemplate.sol
```diff
@@ -48,24 +52,64 @@ contract CompanyTemplate is BaseTemplate {
...
 
     function newInstance(string _id, address[] _holders, uint256[] _stakes, uint64[3] _votingSettings, uint64 _financePeriod, bool _useAgentAsVault) public {
+        _verifyCompanyParameters(_holders, _stakes, _votingSettings);
+        (Kernel dao, ACL acl) = _createDAO();
+        (, Voting voting) = _setupApps(dao, acl, _holders, _stakes, _votingSettings, _financePeriod, _useAgentAsVault);
+        _transferRootPermissionsFromTemplate(dao, voting);
+        _registerID(_id, dao);
+    }
+
+    function newInstance(
```
sohkai: We should add a comment to `newInstanceAndToken()` noting that we can't fit the payroll app in a single block with it (I'm seeing that at 6.9m and the delta for including the Payroll app at 1.1m?)

templates/company/contracts/CompanyTemplate.sol
```diff
@@ -48,24 +52,64 @@ contract CompanyTemplate is BaseTemplate {
...
-        _mintTokens(acl, tokenManager, _holders, _stakes);
+        return (finance, voting);
+    }
 
-        _setupPermissions(dao, acl, agentOrVault, voting, finance, tokenManager, _useAgentAsVault);
+    function _setupPayrollApp(Kernel _dao, ACL _acl, uint256[4] _payrollSettings, Finance _finance, Voting _voting) internal {
+        address denominationToken = _toAddress(_payrollSettings[0]);
+        IFeed priceFeed = IFeed(_toAddress(_payrollSettings[1]));
+        uint64 rateExpiryTime = _payrollSettings[2].toUint64();
+        address employeeManager = _toAddress(_payrollSettings[3]);
```
sohkai: One small thing is perhaps doing this coercion at the start of `newInstance()` instead, so the user doesn't waste gas creating an entire org before their input fails here.

shared/contracts/BaseTemplate.sol
facuspagnuolo: yup

templates/trust/package.json
ajsantander: There's something weird going on, if that dependency isn't there, Trust doesn't compile. Or at least that was the case with the CI.

shared/helpers/assertRevert.js
ajsantander: For some odd reason, extracting the tx and sending it via web3.eth.sendTransaction does not produce a revert error with the correct reason. Sending it via the truffle object does not have that problem.

templates/company/test/company.js
ajsantander: Sure, but beware that it needs to be a contract since `isContract()` is run on that address.

shared/helpers/assertRevert.js
sohkai: Hmm, this is awkward. It reverts with a different reason when sending through with web3?

ajsantander: It reverts with no reason, just a plain "reverted" error.

templates/company/contracts/CompanyTemplate.sol
facuspagnuolo: see my summary comment bellow

facuspagnuolo:
Being wasting a lot of time dealing with `stack too deep` issues... we do have to start thinking on either using ABI encoder V2 or having less optimized templates (e.g. reading settings from storage, using more txs, or sth like that). 

Note that the code looks completely awkward due to how we need to load from storage, pass by argument, and the rest of ticks and tricks that we needed to do to avoid compiling issues, it definitely won't look good for inexperienced users as an example of how to build a template :/

ajsantander:
> I've been wasting a lot of time dealing with `stack too deep` issues... we do have to start thinking on either using ABI encoder V2 or having less optimized templates (e.g. reading settings from storage, using more txs, or sth like that).
> 
> Note that the code looks completely awkward due to how we need to load from storage, pass by argument, and the rest of ticks and tricks that we needed to do to avoid compiling issues, it definitely won't look good for inexperienced users as an example of how to build a template :/

@facuspagnuolo
ABI encoder V2 sounds nice, but we would need to either enable experimental features on contracts (not even sure if 0.4.24 has the feature or how stable it is), or update all contracts to a newer version of Solidity like 0.5.x.

IDEA: What about implementing a very simple system in which any template can be initialized with 2 transactions, no matter how simple or complex it is.

tx 1) The user deploys a "configuration contract" that simply holds state in a bunch of public variables that can be read by later by a template being initialized.

tx 2) The template's initialize function, instead of getting its settings from parameters in a function call, retrieves them from the "configuration contract" by reading state. The template could even selfdestruct the "configuration contract", getting a bit of gas refunded.

templates/company-board/contracts/CompanyBoardTemplate.sol
```diff
@@ -47,116 +50,120 @@ contract CompanyBoardTemplate is BaseTemplate {
...
-        uint64[3] _boardVotingSettings,
-        uint64[3] _shareVotingSettings,
-        uint64 _financePeriod,
-        bool _useAgentAsVault
-    )
-        external
-    {
-        Kernel dao = _popDaoCache(msg.sender);
-        _setupApps(dao, _boardMembers, _shareHolders, _shareStakes, _boardVotingSettings, _shareVotingSettings, _financePeriod, _useAgentAsVault);
+    function setupShare(string _id, address[] _holders, uint256[] _stakes, uint64[3] _votingSettings, uint64 _financePeriod, bool _useAgentAsVault) external {
```
sohkai: If I understand it correctly now, the flow is:

1. `prepareInstance()`
1. `setupBoard()`
1. `setupShare()`

Is it not possible to combine `prepareInstance()` with `setupBoard()`?

templates/trust/package.json
sohkai: Hmm, I see, the contract resolution algorithm is a bit tricky with how npm works (and it's why we can't hoist contracts when using lerna).

Not sure if truffle@5 has made any inroads to fixing this, but for now I've moved the unused apps into the dev dependencies.

sohkai:
> using ABI encoder V2

Would ABI encoder V2 help with the stack issues? There's probably not too much we can do on that regard since we only get 16 slots?

> IDEA: What about implementing a very simple system in which any template can be initialized with 2 transactions, no matter how basic or complex it is.

I'm not sure if this would be cheaper than just passing along arguments; the main benefit I see with deploying a configuration contract (or writing to a global, shared one), would be ease of recovering if you only sent 1 / 3 transactions during onboarding or etc.

sohkai:
@facuspagnuolo I've pushed a few commits that are mostly cosmetic / small fixes. Looking into changing the initialization structure of the company with board template (https://github.com/aragon/dao-templates/pull/124#discussion_r311472629) to see if I can come up with something more intuitive :).

templates/company-board/test/company-board.js
```diff
@@ -21,10 +21,14 @@ const MiniMeToken = artifacts.require('MiniMeToken')
 const PublicResolver = artifacts.require('PublicResolver')
 const EVMScriptRegistry = artifacts.require('EVMScriptRegistry')
 
+const ONE_DAY = 60 * 60 * 24
+const ONE_WEEK = ONE_DAY * 7
+const THIRTY_DAYS = ONE_DAY * 30
+const TWO_MONTHS = ONE_DAY * 31
```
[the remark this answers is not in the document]
facuspagnuolo: Ohhh, it definitely should be `ONE_MONTH`

templates/company-board/contracts/CompanyBoardTemplate.sol
```diff
@@ -47,116 +50,120 @@ contract CompanyBoardTemplate is BaseTemplate {
...
+
+    function _installVaultAndFinance(uint64 _financePeriod, bool _useAgentAsVault) internal returns (Vault agentOrVault, Finance finance) {
+        Kernel dao = _fetchDaoCache();
+        agentOrVault = _useAgentAsVault ? _installDefaultAgentApp(dao) : _installVaultApp(dao);
+        finance = _installFinanceApp(dao, agentOrVault, _financePeriod == 0 ? DEFAULT_FINANCE_PERIOD : _financePeriod);
+    }
+
+    function _setupPermissions(Voting _shareVoting, TokenManager _shareTokenManager, Vault _agentOrVault, Finance _finance, bool _useAgentAsVault) internal returns (Voting) {
+        Kernel dao = _fetchDaoCache();
+        ACL _acl = ACL(dao.acl());
```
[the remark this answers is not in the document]
facuspagnuolo: yup

ajsantander:
I made a quick experiment in remix to see if the configuration contract would help, and it looks like the technique produces stack too deep errors even sooner than when using plain old function parameters. If you want to have a look: https://remix.ethereum.org/#version=soljson-v0.5.1+commit.c8a2cb62.js&optimize=false&gist=17461f2c1dbd9edb598acba1a46ea412

sohkai:
That is definitely not a good sign for the configuration contract... it's probably because doing a call involves quite a lot of stack variables.

templates/company-board/contracts/CompanyBoardTemplate.sol
facuspagnuolo: It may be possible since they cost 5e6 and 1.3e6 respectively, which means they should fit in a single tx

PR #125, opened 2019-08-01
Optimizes app installations

### LINKED ISSUE #105: Optimize apps installation
Current `BaseTemplate` contract installs applications performing 2 transactions, one for the installation per se and another one to initialize the new app. We could leverage the install + initialization function of the `Kernel` to reduce gas costs

ajsantander:
:checkered_flag: **Ready for Review :checkered_flag:  
Fix #105

### DIFF
--- a/shared/contracts/BaseTemplate.sol
+++ b/shared/contracts/BaseTemplate.sol
@@ -32,6 +32,11 @@ contract BaseTemplate is APMNamehash, IsContract {
     bytes32 constant internal FINANCE_APP_ID = 0xbf8491150dafc5dcaee5b861414dca922de09ccffa344964ae167212e8c673ae;
     bytes32 constant internal TOKEN_MANAGER_APP_ID = 0x6b20a3010614eeebf2138ccec99f028a61c811b3b1a3343b6ff635985c75c91f;
 
+    bytes4 constant private AGENT_INIT_SELECTOR = 0x8129fc1c;
+    bytes4 constant private VAULT_INIT_SELECTOR = 0x8129fc1c;
+    bytes4 constant private FINANCE_INIT_SELECTOR = 0x1798de81;
+    bytes4 constant private VOTING_INIT_SELECTOR = 0xdf3d3305;
+
     string constant private ERROR_ENS_NOT_CONTRACT = "TEMPLATE_ENS_NOT_CONTRACT";
     string constant private ERROR_DAO_FACTORY_NOT_CONTRACT = "TEMPLATE_DAO_FAC_NOT_CONTRACT";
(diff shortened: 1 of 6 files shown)


ajsantander:
@facuspagnuolo note how the optimization couldn't be applied to the TokenManager app. This is because the TokenManager's initialize function will fail if it is not the current controller of the token.

This could be overcome by precomputing the address of the yet-to-exist TokenManager instance, since it should be a function of the current contract's address and its nonce. However, since the nonce is not know by a contract, the implementation may not be trivial.

shared/contracts/BaseTemplate.sol
```diff
@@ -32,6 +32,11 @@ contract BaseTemplate is APMNamehash, IsContract {
     bytes32 constant internal FINANCE_APP_ID = 0xbf8491150dafc5dcaee5b861414dca922de09ccffa344964ae167212e8c673ae;
     bytes32 constant internal TOKEN_MANAGER_APP_ID = 0x6b20a3010614eeebf2138ccec99f028a61c811b3b1a3343b6ff635985c75c91f;
 
+    bytes4 constant private AGENT_INIT_SELECTOR = 0x8129fc1c;
+    bytes4 constant private VAULT_INIT_SELECTOR = 0x8129fc1c;
+    bytes4 constant private FINANCE_INIT_SELECTOR = 0x1798de81;
+    bytes4 constant private VOTING_INIT_SELECTOR = 0xdf3d3305;
```
sohkai: Do we need this? Can't we do something like `abi.encodeWithSelector(Agent.initialize.selector)`?

templates/company-board/test/company-board.js
```diff
@@ -183,13 +183,13 @@ contract('Company with board', ([_, owner, boardMember1, boardMember2, shareHold
         shareTokenManager = TokenManager.at(installedApps['token-manager'][1])
       })
 
-      it('costs ~10.4e6 gas', async () => {
+      it('costs max ~10.29e6 gas', async () => {
```
sohkai: Mmm admittedly this is less of an optimization than I had anticipated.

facuspagnuolo: yep, it saves only one call

shared/contracts/BaseTemplate.sol
ajsantander: I tried to go that way, but the contract needs to be instantiated to be able to use `.selector`.

templates/company-board/test/company-board.js
ajsantander: I don't think that it's minor at all. It's about 5k gas per app. The measurements were too coarse.

PR #126, opened 2019-08-01
Adds survey app

### LINKED ISSUE #104: Add helpers to manage Survey app
Following the new `BaseTemplate` helpers, we should add support for the Survey app

ajsantander:
Fix #104


templates/bare/package.json
```diff
@@ -35,6 +35,7 @@
     "@aragon/apps-vault": "^4.1.0",
     "@aragon/apps-voting": "^2.1.0",
     "@aragon/apps-finance": "^3.0.0",
+    "@aragon/apps-survey": "^1.0.0",
```
sohkai: I don't think these other templates need the survey dependency?

shared/contracts/BaseTemplate.sol
```diff
@@ -179,6 +182,20 @@ contract BaseTemplate is APMNamehash, IsContract {
         _removePermissionFromTemplate(_acl, _tokenManager, _tokenManager.MINT_ROLE());
     }
 
+    /* SURVEY */
+
+    function _installSurveyApp(Kernel _dao, MiniMeToken _token, uint64 _minParticipationPct, uint64 _surveyTime) internal returns (Survey) {
+        Survey survey = Survey(_installNonDefaultApp(_dao, SURVEY_APP_ID));
+        survey.initialize(_token, _minParticipationPct, _surveyTime);
```
sohkai: Once #125 is merged, we should use the optimized version too :).

templates/bare/package.json
facuspagnuolo: I used to have problems with this as well, for some reason the templates using the `BaseTemplate` were asking for all the apps being used by the base :/
Will merge and polish it if possible

shared/contracts/BaseTemplate.sol
facuspagnuolo: Merging both.. will provide a third PR to update this :)

templates/bare/package.json
ajsantander: I was experiencing the problem of the templates not compiling if the Survey app was not added. Same thing happened when I was adding the Payroll app.

PR #127, opened 2019-08-03
Template enhacements

facuspagnuolo:
**MERGE AFTER #124**

This PR provides some minor enhancements:
- Simplify voting-related helpers
- Simplify token-manager-related helpers
- Provide `TokenCache` reusable contract
- Automate initialize selectors for apps installations

### DIFF
--- a/shared/contracts/BaseTemplate.sol
+++ b/shared/contracts/BaseTemplate.sol
@@ -151,14 +151,19 @@ contract BaseTemplate is APMNamehash, IsContract {
 
     /* VOTING */
 
+    function _installVotingApp(Kernel _dao, MiniMeToken _token, uint64[3] _votingSettings) internal returns (Voting) {
+        return _installVotingApp(_dao, _token, _votingSettings[0], _votingSettings[1], _votingSettings[2]);
+    }
+
     function _installVotingApp(Kernel _dao, MiniMeToken _token, uint64 _support, uint64 _acceptance, uint64 _duration) internal returns (Voting) {
         bytes memory initializeData = abi.encodeWithSelector(Voting(0).initialize.selector, _token, _support, _acceptance, _duration);
         return Voting(_installNonDefaultApp(_dao, VOTING_APP_ID, initializeData));
     }
 
-    function _createVotingPermissions(ACL _acl, Voting _voting, address _grantee, address _manager) internal {
-        _acl.createPermission(_grantee, _voting, _voting.MODIFY_QUORUM_ROLE(), _manager);
-        _acl.createPermission(_grantee, _voting, _voting.MODIFY_SUPPORT_ROLE(), _manager);
+    function _createVotingPermissions(ACL _acl, Voting _voting, address _settingsGrantee, address _createVotesGrantee, address _manager) internal {
+        _acl.createPermission(_settingsGrantee, _voting, _voting.MODIFY_QUORUM_ROLE(), _manager);
+        _acl.createPermission(_settingsGrantee, _voting, _voting.MODIFY_SUPPORT_ROLE(), _manager);
+        _acl.createPermission(_createVotesGrantee, _voting, _voting.CREATE_VOTES_ROLE(), _manager);
     }
 
     /* SURVEY */
--- a/shared/contracts/BaseTemplate.sol
+++ b/shared/contracts/BaseTemplate.sol
@@ -41,11 +41,6 @@ contract BaseTemplate is APMNamehash, IsContract {
     bytes32 constant internal TOKEN_MANAGER_APP_ID = 0x6b20a3010614eeebf2138ccec99f028a61c811b3b1a3343b6ff635985c75c91f;
     bytes32 constant internal SURVEY_APP_ID = 0x030b2ab880b88e228f2da5a3d19a2a31bc10dbf91fb1143776a6de489389471e;
 
-    bytes4 constant private AGENT_INIT_SELECTOR = 0x8129fc1c;
-    bytes4 constant private VAULT_INIT_SELECTOR = 0x8129fc1c;
-    bytes4 constant private FINANCE_INIT_SELECTOR = 0x1798de81;
-    bytes4 constant private VOTING_INIT_SELECTOR = 0xdf3d3305;
-
     string constant private ERROR_ENS_NOT_CONTRACT = "TEMPLATE_ENS_NOT_CONTRACT";
     string constant private ERROR_DAO_FACTORY_NOT_CONTRACT = "TEMPLATE_DAO_FAC_NOT_CONTRACT";
     string constant private ERROR_ARAGON_ID_NOT_PROVIDED = "TEMPLATE_ARAGON_ID_NOT_PROVIDED";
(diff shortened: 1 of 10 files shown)


shared/contracts/BaseTemplate.sol
```diff
@@ -152,21 +150,26 @@ contract BaseTemplate is APMNamehash, IsContract {
 
     /* VOTING */
 
+    function _installVotingApp(Kernel _dao, MiniMeToken _token, uint64[3] _votingSettings) internal returns (Voting) {
```
sohkai:
```suggestion
    function _installVotingApp(Kernel _dao, MiniMeToken _token, uint64[3] memory _votingSettings) internal returns (Voting) {
```

Just to be a bit more explicit :)

shared/contracts/BaseTemplate.sol
```diff
@@ -272,15 +274,15 @@ contract BaseTemplate is APMNamehash, IsContract {
     /* APPS */
 
     function _installNonDefaultApp(Kernel _dao, bytes32 _appId) internal returns (address) {
-        return _installApp(_dao, _appId, new bytes(0), false);
+        return _installNonDefaultApp(_dao, _appId, new bytes(0));
     }
 
     function _installNonDefaultApp(Kernel _dao, bytes32 _appId, bytes _data) internal returns (address) {
```
sohkai: Could we rename `_data` to `_initializeData` here and elsewhere to make it more clear?

shared/contracts/TokenCache.sol
```diff
@@ -0,0 +1,22 @@
...
+    }
+
+    function _popTokenCache(address _owner) internal returns (MiniMeToken) {
+        require(tokenCache[_owner] != address(0), ERROR_MISSING_TOKEN_CACHE);
+
+        MiniMeToken token = MiniMeToken(tokenCache[_owner]);
+        delete tokenCache[_owner];
+        return token;
+    }
+}
```
sohkai: I'm really starting to think we could move these contracts to something like `@aragon/templates-helpers` or etc. so we can publish them just by themselves.

shared/contracts/BaseTemplate.sol
```diff
@@ -124,15 +119,17 @@ contract BaseTemplate is APMNamehash, IsContract {
     /* AGENT */
 
     function _installDefaultAgentApp(Kernel _dao) internal returns (Agent) {
-        Agent agent = Agent(_installDefaultApp(_dao, AGENT_APP_ID, abi.encodeWithSelector(AGENT_INIT_SELECTOR)));
+        bytes memory initializeData = abi.encodeWithSelector(Agent(0).initialize.selector);
```
ajsantander: :raised_hands:

shared/contracts/BaseTemplate.sol
```diff
@@ -152,21 +150,26 @@ contract BaseTemplate is APMNamehash, IsContract {
...
+        return Voting(_installNonDefaultApp(_dao, VOTING_APP_ID, initializeData));
     }
 
-    function _createVotingPermissions(ACL _acl, Voting _voting, address _grantee, address _manager) internal {
-        _acl.createPermission(_grantee, _voting, _voting.MODIFY_QUORUM_ROLE(), _manager);
-        _acl.createPermission(_grantee, _voting, _voting.MODIFY_SUPPORT_ROLE(), _manager);
+    function _createVotingPermissions(ACL _acl, Voting _voting, address _settingsGrantee, address _createVotesGrantee, address _manager) internal {
+        _acl.createPermission(_settingsGrantee, _voting, _voting.MODIFY_QUORUM_ROLE(), _manager);
+        _acl.createPermission(_settingsGrantee, _voting, _voting.MODIFY_SUPPORT_ROLE(), _manager);
+        _acl.createPermission(_createVotesGrantee, _voting, _voting.CREATE_VOTES_ROLE(), _manager);
```
ajsantander: Nice catch, we had missed this one before.

shared/contracts/BaseTemplate.sol
```diff
@@ -152,21 +150,26 @@ contract BaseTemplate is APMNamehash, IsContract {
...
     }
 
     /* SURVEY */
 
     function _installSurveyApp(Kernel _dao, MiniMeToken _token, uint64 _minParticipationPct, uint64 _surveyTime) internal returns (Survey) {
-        Survey survey = Survey(_installNonDefaultApp(_dao, SURVEY_APP_ID));
-        survey.initialize(_token, _minParticipationPct, _surveyTime);
-        return survey;
+        bytes memory initializeData = abi.encodeWithSelector(Survey(0).initialize.selector, _token, _minParticipationPct, _surveyTime);
+        return Survey(_installNonDefaultApp(_dao, SURVEY_APP_ID));
```
ajsantander: The initializeData doesn't seem to be used here. Should it be:
```return Survey(_installNonDefaultApp(_dao, SURVEY_APP_ID, initializeData));```

File: shared/contracts/BaseTemplate.sol [hunk 7]
```diff
@@ -235,8 +238,7 @@ contract BaseTemplate is APMNamehash, IsContract {
 
     function _createTokenManagerPermissions(ACL _acl, TokenManager _tokenManager, address _grantee, address _manager) internal {
         _acl.createPermission(_grantee, _tokenManager, _tokenManager.MINT_ROLE(), _manager);
-        _acl.createPermission(_grantee, _tokenManager, _tokenManager.ASSIGN_ROLE(), _manager);
-        _acl.createPermission(_grantee, _tokenManager, _tokenManager.REVOKE_VESTINGS_ROLE(), _manager);
```
ajsantander: On what criteria are we assigning and not assigning some of the roles here? The roles related to vesting will be particularly useful when integrating the Presale and/or the Fundraising apps.

shared/contracts/BaseTemplate.sol
ajsantander: +1

templates/company/contracts/CompanyTemplate.sol
```diff
@@ -131,31 +128,8 @@ contract CompanyTemplate is BaseTemplate {
...
-    function _cacheToken(MiniMeToken _token, address _owner) internal {
-        tokenCache[_owner] = _token;
-    }
-
-    function _popTokenCache(address _owner) internal returns (MiniMeToken) {
-        require(tokenCache[_owner] != address(0), ERROR_MISSING_TOKEN_CACHE);
-
-        MiniMeToken token = MiniMeToken(tokenCache[_owner]);
-        delete tokenCache[_owner];
-        return token;
```
ajsantander: :heart_eyes:

shared/contracts/BaseTemplate.sol
facuspagnuolo: I'd change all the arrays passed by argument in that case

shared/contracts/TokenCache.sol
facuspagnuolo: Why not publishing this repo then?

shared/contracts/BaseTemplate.sol
```diff
@@ -152,21 +150,26 @@ contract BaseTemplate is APMNamehash, IsContract {
 
     /* VOTING */
 
+    function _installVotingApp(Kernel _dao, MiniMeToken _token, uint64[3] _votingSettings) internal returns (Voting) {
+        return _installVotingApp(_dao, _token, _votingSettings[0], _votingSettings[1], _votingSettings[2]);
```
ajsantander: Also, previously we were checking that the `_votingSettings` length be correct with a require statement and throwing something like "ERROR_BAD_VOTE_SETTINGS" when it was incorrect. Was this not needed?

shared/contracts/BaseTemplate.sol
facuspagnuolo: This would be the default helper, I basically assumed the most reusable stuff based on the templates we have now

shared/contracts/BaseTemplate.sol
facuspagnuolo: Note that we are already limiting that by `uint64[3]`. However, we are still catching that from the templates side

facuspagnuolo:
@sohkai 

> The one hesitation I have is if the additional "syntax sugar" methods, like _installVotingApp(Kernel, MiniMeToken, uint64[3]), are really worth their cost in contract size.

Remember that this does not increase the bytecode length unless you use those helpers

shared/contracts/TokenCache.sol
sohkai: Hmm yes, ok, maybe it would be better to publish `templates-shared` since it also contains useful scripts and etc.

PR #128, opened 2019-08-08
Restructure Company with Board template's initialization

sohkai:
Refactors how the Company with Board template is initialized, grouping more like-elements together.

To create a Company with Board, the flow is now:

1. Call `prepareInstance()`, creating an incomplete org with its two voting apps installed, and
1. Call `finalizeInstance()`, to finish setting up the org with token holders and other apps.

Orgs at the end of step 1 (`prepareInstance()`) will be viewable in an Aragon client, but will have their permissions locked such that no votes can be created, and **only** the template is able to continue adding apps, changing permissions, or minting tokens. The previous versions also held this property.

It's not obvious from the tests, but the changes result in ~16k gas being saved in an org without Payroll installed, and a ~1k save when Payroll is installed. It also makes it _slightly_ easier to add larger groups of board members or share holders (each cost ~50-100k gas), since they happen in the second step that requires less overall gas.

### DIFF
--- a/templates/company-board/contracts/CompanyBoardTemplate.sol
+++ b/templates/company-board/contracts/CompanyBoardTemplate.sol
@@ -110,7 +110,7 @@ contract CompanyBoardTemplate is BaseTemplate {
         _registerID(_id, address(dao));
     }
 
-    function _finalizeApps(Kernel _dao, address[] _shareHolders, uint256[] _shareStakes, address[] _boardMembers, Voting _shareVoting, Voting _boardVoting) internal {
+    function _finalizeApps(Kernel _dao, address[] memory _shareHolders, uint256[] memory _shareStakes, address[] memory _boardMembers, Voting _shareVoting, Voting _boardVoting) internal {
         (MiniMeToken shareToken, MiniMeToken boardToken) = _popTokenCaches();
 
         // Install
@@ -206,7 +206,7 @@ contract CompanyBoardTemplate is BaseTemplate {
         delete c.boardToken;
     }
 
-    function _ensureFinalizationSettings(address[] memory _shareHolders, uint256[] memory _shareStakes, address[] memory _boardMembers) internal {
+    function _ensureFinalizationSettings(address[] memory _shareHolders, uint256[] memory _shareStakes, address[] memory _boardMembers) private pure {
         require(_shareHolders.length > 0, ERROR_MISSING_SHARE_MEMBERS);
         require(_shareHolders.length == _shareStakes.length, ERROR_BAD_HOLDERS_STAKES_LEN);
         require(_boardMembers.length > 0, ERROR_MISSING_BOARD_MEMBERS);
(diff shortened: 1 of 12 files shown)


templates/company-board/test/company-board.js
```diff
@@ -59,98 +59,90 @@ contract('Company with board', ([_, owner, boardMember1, boardMember2, shareHold
...
+    return template.sendTransaction(encodeCall(finalizeInstanceFn, params, txParams))
   }
 
   context('when the creation fails', () => {
     const FINANCE_PERIOD = 0
     const USE_AGENT_AS_VAULT = true
 
     context('when there was no instance prepared before', () => {
-      it('reverts when no board members are given', async () => {
-        await assertRevert(template.prepareInstance(SHARE_TOKEN_NAME, SHARE_TOKEN_SYMBOL, [], BOARD_VOTING_SETTINGS), 'COMPANY_MISSING_BOARD_MEMBERS')
```
sohkai: I tried testing

```js
await assertRevert(template.prepareInstance(SHARE_TOKEN_NAME, SHARE_TOKEN_SYMBOL, [], BOARD_VOTING_SETTINGS), 'COMPANY_MISSING_BOARD_MEMBERS')
```

But the settings don't seem to work very well (it appears their length is always allocated to the expected amount)?

templates/company-board/contracts/CompanyBoardTemplate.sol
```diff
@@ -42,207 +42,173 @@ contract CompanyBoardTemplate is BaseTemplate {
...
+        boardToken = MiniMeToken(c.boardToken);
+        delete c.shareToken;
+        delete c.boardToken;
     }
 
-    function _ensureCompanySettings(address[] _holders, uint256[] _stakes, uint64[3] _votingSettings) private pure {
-        require(_holders.length > 0, ERROR_MISSING_SHARE_MEMBERS);
-        require(_holders.length == _stakes.length, ERROR_BAD_HOLDERS_STAKES_LEN);
-        require(_votingSettings.length == 3, ERROR_BAD_VOTE_SETTINGS);
+    function _ensureFinalizationSettings(address[] memory _shareHolders, uint256[] memory _shareStakes, address[] memory _boardMembers) internal {
```
sohkai: I've added `memory` to the internal functions passing along arrays; what do you think about doing so for the rest of the templates that use this as well?

templates/company-board/test/company-board.js
facuspagnuolo: Yea, exactly

templates/company-board/contracts/CompanyBoardTemplate.sol
ajsantander: It could help when we migrate to Solidity 0.5.x. If we do it, we should also add it for any parameters of type struct, mapping and string.

sohkai:
I've added the `memory` keyword to public and internal functions in 5ab50e5.

PR #130, opened 2019-08-11
WIP: DAO template engine

ajsantander:
:construction: WIP :construction: 

This is a suggestion for implementing a DAO template engine. An idea by @facuspagnuolo 

It is a CLI program with a set of commands (using commander.js), based on "descriptor" yaml files that outline the structure of a DAO (see [./descriptors/company.yaml](https://github.com/aragon/dao-templates/blob/371af295d1a64db18c74f558236cf6d1c6187ed2/engine/descriptors/company.yaml) for an example of such a file).

The program takes in descriptor files and generates output, _or_ analyses deployed DAOs and generates descriptor files.

A very trivial "permissions" command has been implemented to illustrate the suggested architecture. It simply reads a descriptor file and outputs a mardown table representing the DAOs permissions:

`> dao-template-engine permissions ./descriptors/company.yaml`

Outputs the following markdown

App | Permission | Grantee | Manager
--- | --- | --- | ---
kernel|APP_MANAGER_ROLE|share-voting|share-voting
acl|CREATE_PERMISSIONS_ROLE|share-voting|share-voting
share-voting|CREATE_VOTES_ROLE|share-tm|share-voting
share-voting|MODIFY_QUORUM_ROLE|share-voting|share-voting
share-voting|MODIFY_SUPPORT_ROLE|share-voting|share-voting
agent|EXECUTE_ROLE|share-voting|share-voting
agent|RUN_SCRIPT_ROLE|share-voting|share-voting
agent|TRANSFER_ROLE|finance|share-voting
finance|CREATE_PAYMENTS_ROLE|share-voting|share-voting
finance|EXECUTE_PAYMENTS_ROLE|share-voting|share-voting
finance|MANAGE_PAYMENTS_ROLE|share-voting|share-voting
share-tm|BURN_ROLE|share-voting|share-voting
share-tm|MINT_ROLE|share-voting|share-voting

Other possible commands:
-  A "generate" command that takes a descriptor yaml file and generates either the Solidity source code, or the EVM bytecode for a DAO template.
- A "describe" command that takes the address of a deployed DAO and generates a descriptor file.
- A "validate" command that takes the address of a deployed DAO and a descriptor file, and verifies that the DAO conforms to the provided description, stating deviations when they occur.
- etc...


sohkai:
It's not a big deal, but we've historically used `yargs` for most CLI-related programs. I'm not super super happy with `yargs` though, and if this is not something that will grow too complicated, we can stay with `commander.js` :).

ajsantander:
> It's not a big deal, but we've historically used `yargs` for most CLI-related programs. I'm not super super happy with `yargs` though, and if this is not something that will grow too complicated, we can stay with `commander.js` :).

Sure, if we all agree that this is the way to go (a cli program with multiple commands, which could also expose an API for programmatic usage), I'm happy to swap out commander.ys for yargs.

ajsantander:
Thoughts on how a "generate" command could work:

**Option 1, using EVM scripts**
The descriptor file is traversed and calls to a contract similar to BaseTemplate are encoded into actions, which are then bundled into one or more EVM scripts. This contract's functions would need be public, so that they can be encoded into using Web3 `getData()`. The contract would never be deployed, but used as a source of "recipes" to build EVM scripts. The EVM scripts are then passed into an executor contract, which is then called externally as many times that is needed to initialize the DAO.

**Option 2, generating Solidity source**
The descriptor file is traversed and used to produce the Solidity source code for a template that extends BaseTemplate. The generated source code would contain functions that deploy and initialize the DAO. The generated Solidity source is then compiled, deployed and again, called externally N times for initialization.

**Option 3, generating raw EVM bytecode**
Similar to option 1, but without using EVM scripts. This option would be the most powerful but also the most laborious, because it involves writing a compiler of sorts; from yaml to EVM bytecode. An approach might be to use something similar to the BaseTemplate to obtain and hardcode chunks of bytecode into this pseudo-compiler.

All options involve the difficult challenge of knowing how much can be done in a single transaction, limited by the stack depth of the EVM and the block gas limit. The solution would have to keep track of this, and know when to open a new EVM script, a new function or a new chunk of bytecode, depending on the approach.

lkngtn:
This is awesome! 

I think for now keeping this a separate utility makes sense, but when we get to the generate and describe functionality it would make sense for these to be aragonCLI commands. cc @0xGabi @0x6431346e

lkngtn:
Another possible way to approach this which has some pretty clear UX/Cost tradeoffs, but which significantly reduces implementation complexity would be to create an automation script with atomic actions for the deployment. 

Essentially this could output a shell script that would deploy an org based on the yaml file by executing a series of individual transactions using the CLI's DAO commands. The user would potentially have to manually approve and wait for confirmation of these transactions, and we might need to adjust the CLI outputs to make it easier to get the relevant information like the deployed addresses back to the script.

This definitely doesn't feel like the ideal solution and manually waiting for 20+ transactions to process and confirming each one would be fairly tedious and time consuming, but it might be a much quicker path and would I think address the primary challenge of being able to create custom deployments that are easily replicable without writing code.

0xGabi:
Thanks @ajsantander, this idea is super. I'm Reading it a second time and it have great potencial 

The `generate` command is going to be really useful. I like **Option 2** approach cause it is similar to what we have now with the [Template](https://github.com/aragon/aragon-react-boilerplate/blob/master/contracts/Template.sol) file on the boilerplates

I think the `descriptor yaml` should be part of the content published to aragonPM with each template. Is really powerful to have it setup that way and will allow to use this new scripts directly from the repos. 

I'm in favor of moving this new functionality to the aragonCLI 🙌

ajsantander:
@0xGabi 
> @ajsantander can you further explain how that work?

I have no idea :laughing: 
We'd have to think about this one. Note that **@lkngtn's option 4** doesn't suffer from this problem because it does not have the requirement of minimizing the number of transactions needed to set up an org.

ajsantander:
Closing this issue for now. The topic will be re-addressed after some discussions.

PR #131, opened 2019-08-12
Allow `@aragon/templates-shared` to be published

sohkai:
Adds npm configuration to `shared/` so we can publish it :).


shared/package.json
```diff
@@ -1,8 +1,18 @@
...
   "description": "Shared helpers for Aragon templates",
+  "author": "Aragon Association <legal@aragon.org>",
+  "license": "GPL-3.0-or-later",
+  "files": [
+    "contracts/",
+    "ethereum-data",
+    "helpers/",
+    "lib/",
+    "scripts/",
+    ".env",
```
sohkai: I'm not really sure I understand the need for the `.env` file. I see we use `ETHDATA_LOCAL_PATH` in the docker-compose file, but does that get automatically injected by docker?

facuspagnuolo: We shouldn't add it here definitely. It is used by `dotenv` which we already dropped, so we don't need it here.
I don't see why adding `ethereum-data` and `docker-compose.yml`, these completely custom to our testing task.

sohkai: But we're sharing the testing utils with others; do you think it'd be useful if they had these geth-helpers too (like we do in our templates)?

facuspagnuolo: Now you're asking, yea, it's weird... maybe it is being injected by docker as you were saying

PR #132, opened 2019-08-12
 Add the ability to pass custom apps to publish

osarrouy:
This will make it easier to write custom kits on top of the existing base architecture [like publishing the fundraising apps before running the tests]


shared/scripts/deploy-template.js
```diff
@@ -1,12 +1,13 @@
 const TemplatesDeployer = require('../lib/TemplatesDeployer')
+const { APPS } = require("../helpers/apps");
```
facuspagnuolo: Small suggestion, let's keep the single quotes and no semicolons :)
```suggestion
const { APPS } = require('../helpers/apps')
```

PR #133, opened 2019-08-13
Base: Emit an event when an organization has been finalized

ajsantander:
Kept the original **DeployDao** event when a DAO is created, and added a new **RegisterDao** event to represent the finalization of the template, since the last thing all templates do is call `_registerID(...)`.

The RegisterDao event carries the parameter `string id`, with the ID that the DAO is registered with, and this event + value are checked for on all tests.

### DIFF
--- a/shared/contracts/BaseTemplate.sol
+++ b/shared/contracts/BaseTemplate.sol
@@ -55,6 +55,7 @@ contract BaseTemplate is APMNamehash, IsContract {
     IFIFSResolvingRegistrar internal aragonID;
 
     event DeployDao(address dao);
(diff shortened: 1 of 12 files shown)


ajsantander:
After discussing this with @sohkai, we decided to rename the event to **SetupDao**, that it carries the address of the DAO, and that it is emitted from within `_transferRootPermissionsFromTemplate(...)`.

ajsantander:
Note that if future templates call `_transferPermissionsFromTemplate(...)` instead of `_transferRootPermissionsFromTemplate(...)` or manually revoke permissions from the template, the SetupDao event won't be triggered.

PR #136, opened 2019-08-18
Publish script for aragen && include register-packages option on deployer

0xGabi:


0xGabi:
Sorry about the lint and CLA issues. I restore my pc and still need a few more adjustments.

Btw how do you usually configure your editor when there is no rc file on the repo? 🤔

PR #137, opened 2019-08-18
Publish script for aragen && include register-packages option on deployer

0xGabi:
This PR will help with the automation of aragen releases.

The current deployer script don't took in consideration the case that we don't want to register the packages even if we are in a local devchain. As a consequence when trying to publish to aragonPM with the aragonCLI `aragon apm publish` command we get a VM exception with the error: `revert REPO_INVALID_VERSION`.

The solution I have implemented provide a new option to the deployer, `register-pacakges`, that allow to prevent packages registration during template deployment. 

I also notice that after a packages was registered using the [`_registerPackage`](https://github.com/aragon/dao-templates/blob/master/shared/lib/TemplatesDeployer.js#L161) function it return the following:
```
❯ aragon apm info bare-template
✖ Fetching bare-template.aragonpm.eth@latest
    → The parameter must be a valid HEX string.
```


0xGabi:
Btw how do you usually configure your editor when there is no `.rc` files on the repo? 🤔

templates/bare/package.json
```diff
@@ -17,16 +17,12 @@
...
     "deploy:ropsten": "npm run compile && truffle exec ./scripts/deploy.js --network ropsten --ens 0x6afe2cacee211ea9179992f89dc61ff25c61e923 --dao-factory 0x3f2aa9dd22e97070518ba7988fe9b8724129d497 --mini-me-factory 0x1ce5621d386b2801f5600f1dbe29522805b8ac11",
     "deploy:mainnet": "npm run compile && truffle exec ./scripts/deploy.js --network mainnet --ens 0x314159265dd8dbb310642f98f50c066173c1259b --dao-factory 0xc29f0599df12eb4cbe1a34354c4bac6d944071d1 --mini-me-factory 0x909d05f384d0663ed4be59863815ab43b4f347ec",
-    "publish:rpc": "aragon apm publish major $(npm run deploy:rpc | tail -n 1) --environment default",
-    "publish:devnet": "aragon apm publish major $(npm run deploy:devnet | tail -n 1) --environment default",
-    "publish:staging": "aragon apm publish major $(npm run deploy:rinkeby | tail -n 1) --environment rinkeby",
-    "publish:rinkeby": "aragon apm publish major $(npm run deploy:rinkeby | tail -n 1) --environment rinkeby",
-    "publish:ropsten": "aragon apm publish major $(npm run deploy:ropsten | tail -n 1) --environment ropsten",
-    "publish:mainnet": "aragon apm publish major $(npm run deploy:mainnet | tail -n 1) --environment mainnet"
```
sohkai: We should keep these other publishing scripts.

shared/scripts/deploy-template.js
```diff
@@ -7,23 +7,45 @@ const errorOut = message => {
...
+  contractName,
+  apps = APPS
+) {
   let { ens, owner, verbose, daoFactory, miniMeFactory } = require('yargs')
     .option('e', { alias: 'ens', describe: 'ENS address', type: 'string' })
     .option('o', { alias: 'owner', describe: 'Sender address. Will use first address if no one is given.', type: 'string' })
     .option('v', { alias: 'verbose', describe: 'Verbose mode', type: 'boolean', default: false })
     .option('df', { alias: 'dao-factory', describe: 'DAO Factory address. Will deploy new instance if not given.', type: 'string' })
     .option('mf', { alias: 'mini-me-factory', describe: 'MiniMe Factory address. Will deploy new instance if not given.', type: 'string' })
+    .option('rp', { alias: 'register-packages', describe: 'Whether the script will register the packages to aragon', type: 'boolean', default: true })
```
facuspagnuolo: WDYT about using simply `register`?

shared/scripts/deploy-template.js
```diff
@@ -7,23 +7,45 @@ const errorOut = message => {
...
   if (!contractName) errorOut('Missing template contract name.')
 
-  const deployer = new TemplatesDeployer(web3, artifacts, owner, { apps, ens, verbose, daoFactory, miniMeFactory })
+  const deployer = new TemplatesDeployer(web3, artifacts, owner, {
+    apps,
+    ens,
+    verbose,
+    daoFactory,
+    miniMeFactory,
+  })
```
facuspagnuolo: You're not passing the `register` argument, right?

0xGabi:
Thanks for the feedback Facu. Actually I think it will only prevent to register in case there are no previous version for that package. 

I don't fully understand why you need to register if there is no version. [This line](https://github.com/aragon/dao-templates/blob/master/shared/lib/TemplatesDeployer.js#L163) I'm referring to

shared/lib/TemplatesDeployer.js
```diff
@@ -85,35 +98,56 @@ module.exports = class TemplateDeployer {
...
+          artifacts: this.artifacts,
+          web3: this.web3,
+          owner: this.owner,
+          ensAddress: this.ens.address,
+          verbose: this.verbose,
+        })
         const aragonIDAddress = await this._fetchRegisteredAragonID()
-        if (!aragonIDAddress) this.error('Local aragon ID deployment failed, aborting.')
+        if (!aragonIDAddress)
+          this.error('Local aragon ID deployment failed, aborting.')
```
sohkai: Let's put braces around this if

shared/lib/TemplatesDeployer.js
```diff
@@ -177,13 +219,15 @@ module.exports = class TemplateDeployer {
...
+    )
   }
 
   async _writeArappFile(templateName, template) {
-    const { constructor: { contractName } } = template
-    await this.arapp.write(templateName, contractName, this.ens.address)
-    this.log(`Template addresses saved to ${await this.arapp.filePath()}`)
+  const { constructor: { contractName } } = template
+  await this.arapp.write(templateName, contractName, this.ens.address)
+  this.log(`Template addresses saved to ${await this.arapp.filePath()}`)
```
sohkai: This indentation change seems to be weird

shared/scripts/deploy-template.js
```diff
@@ -7,23 +7,46 @@ const errorOut = message => {
...
     .option('v', { alias: 'verbose', describe: 'Verbose mode', type: 'boolean', default: false })
     .option('df', { alias: 'dao-factory', describe: 'DAO Factory address. Will deploy new instance if not given.', type: 'string' })
     .option('mf', { alias: 'mini-me-factory', describe: 'MiniMe Factory address. Will deploy new instance if not given.', type: 'string' })
+    .option('r', { alias: 'register', describe: 'Whether the script will register the packages to aragon', type: 'boolean', default: true })
     .help('help')
     .parse()
 
-  if (!web3) errorOut('Missing "web3" object. This script must be run with a "web3" object globally defined, for example through "truffle exec".')
-  if (!artifacts) errorOut('Missing "artifacts" object. This script must be run with an "artifacts" object globally defined, for example through "truffle exec".')
+  if (!web3)
```
sohkai: We should put braces for all of these multiline `if`s; they get very confusing very quickly without.

templates/reputation/package.json
```diff
@@ -25,10 +25,12 @@
...
+    "deploy:staging": "npm run compile && truffle exec ./scripts/deploy.js --network rinkeby --ens 0xfe03625ea880a8cba336f9b5ad6e15b0a3b5a939 --dao-factory 0xbca2c99aa7018edcde60e6c5744f68003112535d --mini-me-factory 0xac443f983c38c02149ce2ffcff8b7de90dccf77c",
+    "deploy:rinkeby": "npm run compile && truffle exec ./scripts/deploy.js --network rinkeby --ens 0x98Df287B6C145399Aaa709692c8D308357bC085D --dao-factory 0xfdef49fbfe37704af55636bdd4b6bc8cd19143f6 --mini-me-factory 0x6ffeb4038f7f077c4d20eaf1706980caec31e2bf",
+    "deploy:ropsten": "npm run compile && truffle exec ./scripts/deploy.js --network ropsten --ens 0x6afe2cacee211ea9179992f89dc61ff25c61e923 --dao-factory 0x3f2aa9dd22e97070518ba7988fe9b8724129d497 --mini-me-factory 0x1ce5621d386b2801f5600f1dbe29522805b8ac11",
+    "deploy:mainnet": "npm run compile && truffle exec ./scripts/deploy.js --network mainnet --ens 0x314159265dd8dbb310642f98f50c066173c1259b --dao-factory 0xc29f0599df12eb4cbe1a34354c4bac6d944071d1 --mini-me-factory 0x909d05f384d0663ed4be59863815ab43b4f347ec",
+    "publish:aragen": "aragon apm publish major $(npm run deploy:aragen | tail -n 1) --environment default",
```
sohkai: Perhaps we can move the `publish:aragen` command to be just before the `publish:staging` command in all of these templates?

These have been ordered in terms of "network severity" (i.e. `rpc -> devnet -> staging -> rinkeby -> ropsten -> mainnet`)

PR #138, opened 2019-08-21
Added a permissionsTable script for yaml descriptors

ajsantander:
This PR replaces the too ambitious "dao engine" PR https://github.com/aragon/dao-templates/pull/130

It simply adds a "descriptor" package which contains scripts that can be run via npm. For now, it includes a "permissionsTable" script that can convert a yaml descriptor file into a permissions markdown table.

### DIFF
--- a/descriptor/descriptors/company.yaml
+++ b/descriptor/descriptors/company.yaml
@@ -1,0 +1,84 @@
+---
+id: company
+tokens:
+- ref: share
+  symbol: Share
+  name: Share Token
+  decimals: 18
+apps:
+- ref: shareholder-voting
+  type: voting
+  token: share
+  support: 50
+  acceptance: 5
+  duration: 604800
+- ref: share-tm
+  type: token-manager
+  token: share
+- type: finance
+  default: true
+  vault: agent
+  period: 2592000
+- ref: agent
+  type: agent
+  default: true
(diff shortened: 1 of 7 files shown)


descriptor/src/scripts/permissionsTable.js
```diff
@@ -0,0 +1,18 @@
+const readYaml = require('../utils/readYaml')
+
+const args = process.argv.slice(2)
+const descriptorFilePath = args[0]
+
+const yaml = readYaml(descriptorFilePath)
```
sohkai: It would be nice to do a bit of sanity checking here to catch that the path exists, and can be read from, and if not, warn the user.

descriptor/descriptors/company.yaml
```diff
@@ -0,0 +1,84 @@
+---
```
sohkai: Perhaps we can have this in an `examples/` folder instead?

descriptor/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "aragon-template-descriptor-util",
```
sohkai:
```suggestion
  "name": "@aragon/template-descriptor",
```

And we might as well add the "private: true" declaration here too for now.

PR #139, opened 2019-08-21
Removed unused mock contracts

ajsantander:
Also removed the increaseTime.js helper, since it uses one of these contracts and doesn't seem to be used itself either.


sohkai:
I think we use the `ExecutionTarget` as a mock for the PPF instances for Payroll in each of the template's tests, but we can change that to use any other contract instead.

ajsantander:
@sohkai you're right, some tests use ExecutionTarget, but it seems like only for obtaining an address that contains deployed bytecode (apparently for isContract). There are tests in dao-templates that use it, like the company template's tests: https://github.com/aragon/dao-templates/blob/master/templates/company/test/company.js#L24

These tests now use Migrations for this purpose.

PR #145, opened 2019-08-29
Audit review 6.6: Missing memory keyword

ajsantander:
#### Problem:
There was a missing memory keyword in ReputationTemplate L128.

#### Fix:
Added it! At first, it seemed that other memory keywords were missing, but it can't be used in `external` functions, so they're correct in just reporting this single one.

#### Audit issue:
https://github.com/aragonone/aragon-daotemplates-audit-report-2019-08#66-reputation---missing-data-location-for-argument


sohkai:
> it can't be used in external functions

Those are `calldata` instead :).

PR #148, opened 2019-08-29
Audit review 7.2: Improve linting

ajsantander:
#### Problem:
Direct runs of solium and solhint report some linting errors in contracts.

#### Solution:
Added linting to the shared package, some missing .soliumignore files in the template packages, and modified `indentation` and `error-reason` rules in all .soliumrc.json files.

#### Audit issue:
https://github.com/aragonone/aragon-daotemplates-audit-report-2019-08#72-ethlint


shared/contracts/BaseTemplate.sol
```diff
@@ -210,7 +248,11 @@ contract BaseTemplate is APMNamehash, IsContract {
         _acl.createPermission(_settingsManager, _payroll, _payroll.MANAGE_ALLOWED_TOKENS_ROLE(), _permissionsManager);
     }
 
-    function _unwrapPayrollSettings(uint256[4] memory _payrollSettings) internal pure returns (address denominationToken, IFeed priceFeed, uint64 rateExpiryTime, address employeeManager) {
+    function _unwrapPayrollSettings(
+        uint256[4] memory _payrollSettings
+    )
```
facuspagnuolo: are these kind of breakdowns necessary?

ajsantander: Yeah, to fulfill the rule of not having lines with more than 120 chars. We could of course remove the rule instead.

sohkai: Perhaps we can set it to something else. I'm pretty sure we don't follow this rule in other contract repos (since a lot of them have lines that go over 120 char), but can't find the configuration.

PR #157, opened 2019-09-04
Chore: Fix templates deployer to avoid writing

facuspagnuolo:


templates/reputation/arapp.json
```diff
@@ -26,10 +26,11 @@
     "staging": {
       "appName": "reputation-template.aragonpm.eth",
       "network": "rinkeby",
+      "address": "0xe12ac6d33eb6b474524f57e89ee219bcec7d086c",
```
sohkai: I don't think these `address` fields are needed since we can do the ENS lookup.

facuspagnuolo:
True, this is being done by the templates deployer, I think I carried this error since the refactor. Will turn this PR into a fix for that then

PR #177, opened 2021-09-29
update: added support for Harmony One Blockchain

victaphu:
Updated the arapp.json files for the different templates in the DAO.
Note: registry has only been deployed to the TESTNET. I have left the mainnet deployment blank for the time being (need to merge with an ENS-like smart contract)

Note update to package.json (this was required as the code-base is using an older version of node)


victaphu:
I'll redo this pull, the CLA is not signed and the user zain is a misconfigured user on my system

